PR #122, opened 2026-03-24
fix: bump version to 0.2.0 and add version/tag validation

### LINKED ISSUE #94: [Bug] Binary reports 0.1.0 despite being from v0.2.0 release
## Problem

The install script downloads v0.2.0 but the binary reports v0.1.0:

\`\`\`
$ curl -fsSL .../install.sh | sh
Latest version: v0.2.0
Downloading .../syfrah-v0.2.0-x86_64-unknown-linux-musl.tar.gz...
Installing syfrah to /usr/local/bin...
Verifying installation...
syfrah 0.1.0    ← WRONG

$ syfrah --version
syfrah 0.1.0    ← WRONG
\`\`\`

The binary is definitely from the v0.2.0 release (the URL and archive name confirm it), but \`--version\` prints \`0.1.0\`.

### Root cause

\`syfrah --version\` uses the version compiled into the binary via \`clap\`:

\`\`\`rust
// bin/syfrah/src/main.rs
#[derive(Parser)]
#[command(version)]  // ← reads from Cargo.toml at compile time
struct Cli { ... }
\`\`\`

This resolves to \`env!("CARGO_PKG_VERSION")\` which is baked in at build time from \`Cargo.toml\`.

**The bug:** When the v0.2.0 release was built, \`Cargo.toml\` still had \`version = "0.1.0"\`. The version in Cargo.toml was never bumped to \`0.2.0\` before building the release.

### The version chain that must be consistent

\`\`\`
Cargo.toml version  →  compiled into binary  →  --version output
       ↓
   git tag name  →  GitHub Release name  →  install.sh "Latest version"
       ↓
   archive filename (syfrah-v0.2.0-...)
\`\`\`

Currently only the tag/release/filename say \`0.2.0\`. The binary says \`0.1.0\` because Cargo.toml was never updated.

### How this happened

The release workflow creates a tag and builds from the code at that commit. If Cargo.toml wasn't bumped before tagging:
1. Developer tags \`v0.2.0\`
2. CI builds from that commit
3. Cargo.toml says \`0.1.0\`
4. Binary is compiled with \`0.1.0\` baked in
5. Binary is uploaded to a release named \`v0.2.0\`
6. Mismatch

### 1. Bump Cargo.toml BEFORE tagging

The release process must enforce: **Cargo.toml version == git tag**. This is a process issue, not a code issue.

### 2. CI validation: fail if version doesn't match tag

Add a check in the release workflow:

\`\`\`yaml
- name: Verify version matches tag
  run: |
    CARGO_VERSION=\$(grep '^version' Cargo.toml | head -1 | sed 's/.*"\(.*\)".*/\1/')
    TAG_VERSION=\${{ github.ref_name }}  # e.g., v0.2.0
    TAG_VERSION=\${TAG_VERSION#v}        # strip 'v' prefix → 0.2.0
    if [ "\$CARGO_VERSION" != "\$TAG_VERSION" ]; then
      echo "ERROR: Cargo.toml version (\$CARGO_VERSION) != tag (\$TAG_VERSION)"
      exit 1
    fi
\`\`\`

This prevents the mismatch from ever happening again.

### 3. Fix the current release

- Bump Cargo.toml to \`0.2.0\`
- Rebuild the binaries
- Re-upload to the v0.2.0 release (replace existing assets)
- Verify: \`syfrah --version\` outputs \`syfrah 0.2.0\`

### 4. Auto-release (#78) prevents this

If #78 is implemented (version from Cargo.toml → auto-tag → auto-release), this class of bug becomes impossible. The tag is derived from Cargo.toml, not the other way around.

## Done when

- [ ] v0.2.0 release assets rebuilt with correct version in binary
- [ ] \`syfrah --version\` matches the release tag
- [ ] Release workflow includes version/tag match validation (fails if mismatch)
- [ ] Document in release process: "bump Cargo.toml version BEFORE tagging"
- [ ] install.sh verification step checks \`syfrah --version\` matches expected version

sacha-ops:
## Summary

- Bumps workspace version in `Cargo.toml` from `0.1.0` to `0.2.0` so `syfrah --version` reports the correct version
- Adds a version validation step to the release workflow that fails the build if `Cargo.toml` version doesn't match the computed release version
- Updates `install.sh` to verify the installed binary reports the expected version (exits non-zero on mismatch)
- Updates `handbook/releasing.md` to document the version consistency requirement

## Test plan

- [ ] CI passes (fmt, clippy, test)
- [ ] Rebuild v0.2.0 release assets — verify `syfrah --version` outputs `syfrah 0.2.0`
- [ ] Trigger a release with a mismatched Cargo.toml version — verify the build fails at the validation step
- [ ] Run `install.sh` against the rebuilt release — verify version check passes

Closes #94

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,6 +85,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Verify Cargo.toml version matches release version
+        env:
+          VERSION: ${{ needs.version.outputs.next }}
+        run: |
+          CARGO_VERSION=$(grep '^version' Cargo.toml | head -1 | sed 's/.*"\(.*\)".*/\1/')
+          if [ "$CARGO_VERSION" != "$VERSION" ]; then
+            echo "ERROR: Cargo.toml version ($CARGO_VERSION) != release version ($VERSION)"
+            echo "Bump Cargo.toml version to $VERSION before releasing."
+            exit 1
+          fi
+
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: ${{ matrix.target }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.1.0"
+version = "0.2.0"
 edition = "2021"
 license = "Apache-2.0"
 repository = "https://github.com/sifrah/syfrah"
--- a/handbook/releasing.md
+++ b/handbook/releasing.md
@@ -20,10 +20,15 @@ Every merge to `main` automatically produces a new GitHub Release. There is no m
    | `[Feature]` or `feat:` | Minor (0.X.0) | `0.2.0 -> 0.3.0` |
    | Anything else | Patch (0.0.X) | `0.2.0 -> 0.2.1` |
 
-4. **CI gate** -- the workflow waits for CI checks to pass before proceeding.
-5. **Cargo.toml bump** -- the workflow updates `version` in `[workspace.package]` in the root `Cargo.toml`, commits with `release: vX.Y.Z [skip ci]`, and pushes directly to `main`.
-6. **Build** -- all four targets are built in parallel (same matrix as CI).
-7. **Release** -- binaries are packaged into `.tar.gz` archives, `SHA256SUMS.txt` is generated, a git tag `vX.Y.Z` is created, and a GitHub Release is published with all artifacts.
+4. **Version validation** -- each build job verifies that the version in `Cargo.toml` matches the calculated release version. The build fails immediately if they differ.
+5. **Build** -- all four targets are built in parallel (same matrix as CI).
+6. **Release** -- binaries are packaged into `.tar.gz` archives, `SHA256SUMS.txt` is generated, a git tag `vX.Y.Z` is created, and a GitHub Release is published with all artifacts.
+
+### Keeping Cargo.toml in sync
+
+The binary version reported by `syfrah --version` comes from `CARGO_PKG_VERSION`, which is baked in at compile time from `Cargo.toml`. The release workflow validates that `Cargo.toml` matches the computed release version and **will fail the build if they differ**.
+
+**Before merging a version-bumping PR**, update `version` in `[workspace.package]` in the root `Cargo.toml` to match the version that the release workflow will compute. All crates inherit this version via `version.workspace = true`.
 
 ### Influencing the version bump
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -55,8 +55,14 @@ install -m 755 "${TMPDIR}/${BIN}" "${INSTALL_DIR}/${BIN}"
 
 # --- Verify ---
 echo "Verifying installation..."
+EXPECTED_VERSION="${VERSION#v}"
 if command -v "$BIN" > /dev/null 2>&1; then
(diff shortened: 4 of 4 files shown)


sacha-ops (commented):
## Request Changes

**The version validation is unnecessary and will break future releases.**

**What's good**: The intent is sound — version consistency matters, and the install.sh verification is a nice defense-in-depth touch. Documentation updates are welcome.

**Issues**:

1. **`build.rs` already solves this** (`bin/syfrah/build.rs:3-4`): The build script overrides `CARGO_PKG_VERSION` with the `SYFRAH_VERSION` env var, and the release workflow already sets `SYFRAH_VERSION` to the computed version (release.yml:107,113). So release binaries *already* report the correct version regardless of what `Cargo.toml` says. The PR's premise — that Cargo.toml must match for `syfrah --version` to be correct — is a misunderstanding of the existing build pipeline.

2. **The validation step will break the next release** (release.yml:88-96): After this PR, Cargo.toml says `0.2.0`. The next patch merge computes `0.2.1`. The validation fails because `0.2.0 != 0.2.1`. Every single merge to main would require a manual Cargo.toml bump *before* merging, which defeats the automated release design.

3. **Documentation contradiction** (handbook/releasing.md:9): The file still says *"There is no manual version bump required"* but the new "Keeping Cargo.toml in sync" section says the opposite — you must manually update Cargo.toml before merging. These can't both be true.

4. **Fragile version parsing** (release.yml:91): `grep '^version' Cargo.toml | head -1 | sed ...` will break if Cargo.toml gains any other `version` key before the workspace one. `cargo metadata --no-deps --format-version=1 | jq -r '.packages[0].version'` would be robust, but this is moot given issue #1.

**What I'd suggest instead**: Remove the validation step from release.yml entirely — the `build.rs` override already handles release builds. If you want Cargo.toml to reflect the *current* development version (for local `cargo build`), that's a valid goal, but the right mechanism is the automated Cargo.toml bump commit that the old docs described (step 5), not a gate that blocks builds. The install.sh version check is fine to keep.

sacha-ops:
## Review — Soren

**Verdict: APPROVE**

Clean fix for a real problem. The approach is explicit and simple — I respect that.

### What works

- Version bump in `Cargo.toml` is the correct fix. One line, done.
- The CI validation step is a good guardrail. Fail fast, clear error message.
- The `install.sh` version check is a nice touch — catches mismatches at the consumer boundary too.
- Docs updated to match reality. The old docs described an auto-bump step that never existed in the workflow. Good cleanup.

### Minor observations (non-blocking)

1. **`grep '^version' Cargo.toml | head -1 | sed ...`** — This works today because the workspace root has exactly one `^version` line. A slightly more robust alternative would be `cargo metadata --no-deps --format-version 1 | jq -r '.packages[0].version'`, but that requires cargo to be installed before the check runs (it isn't yet at that point in the workflow). The grep approach is fine for now.

2. **Manual sync is inherently fragile.** You need to predict what version the workflow will compute before merging. If someone lands a `feat:` commit and forgets to bump Cargo.toml to the minor version, the release fails. This is the right tradeoff (explicit > magic), but worth being aware of. A future improvement could be a pre-merge CI check that computes the expected version and compares it — but that's out of scope here.

3. **`install.sh` uses `awk '{print $2}'`** — assumes `syfrah --version` outputs `syfrah X.Y.Z`. If the format ever changes, this breaks silently. Acceptable for now.

No over-engineering, no unnecessary complexity. Ship it.

PR #128, opened 2026-03-24
fix: remove version validation that blocks automated releases

### LINKED ISSUE #127: [Bug] Release workflow version validation blocks automated releases
## Problem

The release workflow fails because the version validation step (added in PR #122) requires `Cargo.toml` version to match the computed release version. But the previous automated release system was bumping the version **automatically** based on commit messages — the developer never had to touch `Cargo.toml` manually.

PR #122 added a validation step but broke the auto-bump mechanism. Now every release fails:

```
ERROR: Cargo.toml version (0.2.0) != release version (0.2.13)
Bump Cargo.toml version to 0.2.13 before releasing.
```

## How it worked before

1. Merge to `main` triggers release workflow
2. Workflow inspects commit messages since last tag (`feat:` = minor, `BREAKING` = major, else patch)
3. Workflow **auto-bumps** `Cargo.toml` version, commits with `release: vX.Y.Z [skip ci]`
4. Builds binaries and creates GitHub Release

## What's broken now

The validation step added by PR #122 runs **before** the auto-bump, comparing the old `Cargo.toml` version against the new computed version. It will always fail because the bump hasn't happened yet.

## Expected behavior

The release workflow should either:
- Run the validation **after** the auto-bump commit, not before
- Or remove the validation and trust the auto-bump mechanism (which was working correctly before)

## Acceptance criteria

- [ ] Merging to `main` produces a successful release again
- [ ] `syfrah --version` matches the release tag
- [ ] No manual `Cargo.toml` bump required (automated as before)

sacha-ops:
Closes #127

## Summary

- Removes the `Verify Cargo.toml version matches release version` step from the release workflow
- This step (added in PR #122) ran **before** the auto-bump, comparing the stale `Cargo.toml` version against the computed next version — guaranteed to fail
- The `SYFRAH_VERSION` env var already overrides `CARGO_PKG_VERSION` at build time via `build.rs`, so `syfrah --version` correctly reports the release tag without needing `Cargo.toml` to match

## Test plan

- [ ] Merge to `main` and verify the release workflow completes successfully
- [ ] Confirm `syfrah --version` on the built binary matches the release tag

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,17 +85,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Verify Cargo.toml version matches release version
-        env:
-          VERSION: ${{ needs.version.outputs.next }}
-        run: |
-          CARGO_VERSION=$(grep '^version' Cargo.toml | head -1 | sed 's/.*"\(.*\)".*/\1/')
-          if [ "$CARGO_VERSION" != "$VERSION" ]; then
-            echo "ERROR: Cargo.toml version ($CARGO_VERSION) != release version ($VERSION)"
-            echo "Bump Cargo.toml version to $VERSION before releasing."
-            exit 1
-          fi
-
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: ${{ matrix.target }}


sacha-ops:
## Review — Ren

**Verdict: APPROVE**

The validation step at `release.yml:88-97` was dead on arrival. The `version` job auto-computes `next` from git tags and commit history, but nothing bumps `Cargo.toml` before the check runs — so `$CARGO_VERSION != $VERSION` is guaranteed to be true on every release after a tag increment. This is a classic "test that always fails" situation, which is worse than no test at all because it blocks the entire pipeline.

The `build.rs:3-4` override via `SYFRAH_VERSION` env var (set at `release.yml:118` and `release.yml:124`) already ensures the binary reports the correct version. The safety net the removed step was trying to provide is already handled at a lower, more reliable layer.

One thing to keep an eye on: `Cargo.toml` version will now drift from release tags indefinitely. That's fine for the binary (build.rs handles it), but it could confuse anyone running `cargo install` from source or inspecting `Cargo.toml` directly. Not a blocker — just a known trade-off worth documenting if it comes up. If you ever want a belt-and-suspenders check, the right place would be *after* the version job auto-bumps Cargo.toml (which doesn't exist yet), not before the build.

Clean removal, correct reasoning. Ship it.

sacha-ops:
## Review — Soren

**Verdict: APPROVE**

Correct fix. The removed step compared `Cargo.toml`'s static version against the auto-computed next version — a check that could never pass, since nothing bumps `Cargo.toml` before the build.

The version override path is sound: `SYFRAH_VERSION` is set in both build steps (`.github/workflows/release.yml:107`, `:113`) and `bin/syfrah/build.rs:3-4` injects it as `CARGO_PKG_VERSION` at compile time. No validation gap.

Clean removal, nothing else to flag.

PR #163, opened 2026-03-24
fix: replace hardcoded 'region-1' default with 'default'

### LINKED ISSUE #137: [Bug] Hardcoded default region 'region-1' in daemon.rs
## Problem
`daemon.rs` hardcodes `"region-1"` as default region in 4+ places. A node started without `--region` silently gets `region-1`, which is misleading — the operator thinks they haven't set a region but the node reports one.

## Location
`layers/fabric/src/daemon.rs` — search for `region-1`.

## Fix
When no region is specified:
- Either require it explicitly (error if missing)
- Or use a clearly-default value like `"default"` and warn the operator

## Acceptance criteria
- [ ] No hardcoded `region-1` in daemon.rs
- [ ] Default behavior is clear to the operator
- [ ] CLI help text documents the default
- [ ] CI passes

sacha-ops:
Closes #137

## Summary

- Replaced all hardcoded `"region-1"` fallbacks with `"default"` in `daemon.rs` (5 places) and `peering.rs` (1 place)
- Added a `ui::warn()` when no `--region` is specified during `init` or `join`, so the operator knows the default is being used and how to set it explicitly
- Updated CLI help text for `--region` on both `init` and `join` to document the default value
- Updated e2e test expectations to match the new default

## Test plan

- [x] `cargo fmt` — no changes needed
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [ ] Verify e2e scenarios 28, 33, 34, 35 pass with new default region


sacha-ops:
## Review — Ren

**Verdict: REQUEST_CHANGES**

The mechanical `region-1` → `default` swap is clean and the constant extraction is good. But the tests added here don't actually test production code, and there's a hardcoded value that contradicts the rest of the change.

### 1. `auto_init` hardcodes `"default-zone-1"` — inconsistent with every other path

`daemon.rs:165`:
```rust
zone: Some("default-zone-1".to_string()),
```

Every other call site generates the zone via `store::generate_zone(&region, &[])`, which returns `"zone-1"`. The `auto_init` path now produces `"default-zone-1"` — a legacy-format zone name baked in for the new default region. This looks like a leftover from a find-and-replace of `"region-1-zone-1"` → `"default-zone-1"` instead of calling `generate_zone`.

**Fix:** Use `store::generate_zone(DEFAULT_REGION, &[])` here (or just `"zone-1"` if you want a literal), so every code path is consistent.

### 2. Tests don't exercise production code

`daemon.rs:1571-1595` — the three new tests call a test-local `resolve_region()` helper, not any real function. They prove that `Option::unwrap_or_else` works, not that the six production call sites use `DEFAULT_REGION` correctly. If someone adds a seventh call site and fat-fingers it, none of these tests catch it.

**Suggestion:** Either:
- Write a test that calls `setup_init` (or a meaningful subset) with `config.region = None` and asserts the resulting `NodeState` has `region == "default"` and a zone matching `"zone-1"`, or
- At minimum, add a `grep`/`const`-based compile-time assertion that `DEFAULT_REGION` is `"default"` — but the integration-level test is far more valuable.

If `setup_init` is hard to call in a unit test (it touches WireGuard), consider extracting the region/zone resolution into a small pure function and testing that directly — but make sure it's the *actual function used in production*, not a copy pasted into `#[cfg(test)]`.

### 3. Minor: no test for the `auto_init` path at all

`auto_init` (`daemon.rs:152-167`) now uses `DEFAULT_REGION` for region but hardcodes `"default-zone-1"` for zone (see issue 1). There is no test covering this path, and it's the one that got the value wrong.

### 4. e2e scenario 33 uses `grep -q "default"` — fragile match

`tests/e2e/scenarios/33_fabric_zones_peers_display.sh:48`:
```bash
if echo "$output" | grep "node-2" | grep -q "default"; then
```

`"default"` is a very common substring. If any other field on that line contains the word "default", this passes for the wrong reason. The old pattern `"region-1"` was more distinctive. Consider anchoring: `grep -q "default "` (trailing space) or matching the column more precisely.

---

The constant and the warning are good additions. Fix the `auto_init` zone inconsistency and replace the test-local-helper tests with something that exercises real code paths, and this is good to go.

PR #164, opened 2026-03-24
fix: use monotonic clock in replay guard

### LINKED ISSUE #136: [Bug] Replay guard uses SystemTime instead of Instant
## Problem
`peering.rs` uses `SystemTime` for the announce dedup window and replay protection. SystemTime can jump backwards (NTP correction, manual clock change), which could allow replayed announcements to be accepted or valid announcements to be rejected.

## Location
`layers/fabric/src/peering.rs` — replay guard and dedup window.

## Fix
Use `tokio::time::Instant` (monotonic clock) for all timing comparisons. Keep SystemTime only for display/logging purposes.

## Acceptance criteria
- [ ] Replay guard uses Instant for dedup window
- [ ] Announce age check uses Instant
- [ ] Clock jumps don't affect security
- [ ] CI passes

sacha-ops:
Closes #136

## Summary
- **Dedup window** now uses `std::time::Instant` instead of epoch seconds — immune to wall-clock jumps
- **Stale-timestamp check** uses a monotonic epoch anchor (`Instant` + epoch pair captured at `ReplayGuard` construction) so NTP corrections cannot bypass replay protection
- `epoch_now()` (SystemTime) retained only for wire timestamps and display/logging
- All existing tests updated and passing

## Test plan
- [x] `cargo fmt` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — zero warnings
- [x] `cargo test` — all 170+ tests pass
- [x] Replay guard tests exercise stale/fresh/boundary timestamps via monotonic epoch

### DIFF
--- a/layers/fabric/src/peering.rs
+++ b/layers/fabric/src/peering.rs
@@ -948,6 +948,7 @@ pub fn generate_pin() -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::daemon::now;
 
     #[tokio::test]
     async fn replay_guard_detects_duplicate_ciphertext() {


sacha-ops:
## Review — Soren

**Verdict: APPROVE**

Clean, focused change. Monotonic `Instant` for dedup, anchor pair for stale-check — correct approach, minimal surface area.

A few notes (none blocking):

**`peering.rs:90` — `checked_duration_since` semantics are correct but subtle.** `None` (future timestamp) → retain. Good. Worth knowing this relies on `is_none_or` (stable since Rust 1.82).

**`peering.rs:70-71` — non-atomic anchor capture.** `Instant::now()` and `epoch_now()` are two calls; a scheduling delay between them means `epoch_anchor` could be slightly ahead of `instant_anchor`. In practice this makes `monotonic_epoch()` marginally conservative (returns a value ≤ wall-clock), which is the safe direction for replay protection. Fine as-is.

**`peering.rs:77` — `monotonic_epoch` is `pub(crate)` but only used within this module.** Could be `fn` (private). Trivial.

Tests are solid — the eviction test (`peering.rs:1002-1020`) with `checked_sub` backdating is a nice touch. The `monotonic_epoch_never_decreases` test with a 10ms sleep is a reasonable smoke test.

Ship it.

PR #166, opened 2026-03-24
Validate WG key before processing decrypted peer announces

### LINKED ISSUE #141: [Bug] No WG key validation before decrypt in peering.rs
## Problem
`peering.rs` attempts to decrypt peer announcements without validating the sender's WireGuard key format first. A malformed key could cause a panic or unexpected behavior in the crypto layer.

## Fix
Validate the WG key (44-char base64, decodes to 32 bytes) before attempting decryption.

## Acceptance criteria
- [ ] WG key validated before any crypto operation
- [ ] Malformed keys rejected with clear log message
- [ ] CI passes

sacha-ops:
Closes #141

## Summary

- Moved `validate_peer_record()` (which includes WG public key format validation: 44-char base64, decodes to 32 bytes) to run **immediately after** `decrypt_record()` in the `PeerAnnounce` handler
- Previously, decrypted record fields (`last_seen`, `name`) were accessed before validation, allowing malformed WG keys and other invalid fields to reach downstream logic
- Malformed keys are now rejected with a clear warning log before any further processing

## Test plan

- [x] `cargo fmt` — no changes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean

### DIFF
--- a/layers/fabric/src/peering.rs
+++ b/layers/fabric/src/peering.rs
@@ -649,9 +649,7 @@ async fn handle_incoming(
 
             let record = decrypt_record(&ciphertext, &enc_key)?;
 
-            // Validate record fields (including WG key format) before any
-            // further processing to prevent malformed data from reaching
-            // crypto operations or downstream consumers.
+            // Validate before any further processing (field reads, stale check, etc.).
             if let Err(e) = validate_peer_record(&record) {
                 warn!(from = %peer_addr, error = %e, "rejecting peer announce: validation failed");
                 return Err(PeeringError::Protocol(format!(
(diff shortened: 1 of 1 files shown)


sacha-ops:
## Review — Soren

**Verdict: APPROVE**

Clean, minimal change. Moving `validate_peer_record()` above the stale check at `peering.rs:653` is the right call — invalid records should never reach downstream logic, even if they'd be discarded by staleness. The previous ordering let malformed keys silently pass through via `return Ok(())` on the stale path, which is the exact bug #141 describes.

The test at `peering.rs:975` is well-constructed: it crafts a record that is *both* invalid and stale, which precisely targets the ordering regression. If validation ever drifts below the stale check again, this test catches it.

No notes.

PR #167, opened 2026-03-24
fix: add read timeout to control socket connections

### LINKED ISSUE #142: [Bug] Control socket has no read timeout
## Problem
`control.rs` has no timeout on socket accept/read. A slow or malicious client can hold the socket indefinitely, blocking CLI commands from reaching the daemon.

## Fix
Add a read timeout (e.g. 5 seconds) on the control socket. Drop connections that don't send a complete message within the timeout.

## Acceptance criteria
- [ ] Read timeout on control socket connections
- [ ] Slow clients are disconnected with a log warning
- [ ] Normal CLI operations unaffected
- [ ] CI passes

sacha-ops:
Closes #142

## Summary
- Add a 5-second read timeout (`CONTROL_READ_TIMEOUT`) on `handle_control_connection` using `tokio::time::timeout`
- Slow or malicious clients that don't send a complete request within the timeout are dropped with a `warn!` log
- Normal CLI operations are unaffected (typical control messages complete in milliseconds)

## Test plan
- [x] New test `control_slow_client_times_out`: connects to the control socket but sends nothing, verifies the server returns a timeout error within the deadline and the handler is never invoked
- [x] All existing tests pass (clippy clean, no regressions)

### DIFF
--- a/layers/fabric/Cargo.toml
+++ b/layers/fabric/Cargo.toml
@@ -31,3 +31,4 @@ toml = "0.8"
 
 [dev-dependencies]
 tempfile = "3"
+tokio = { workspace = true, features = ["test-util"] }
--- a/layers/fabric/src/control.rs
+++ b/layers/fabric/src/control.rs
@@ -276,6 +276,8 @@ mod tests {
 
     #[tokio::test]
     async fn control_slow_client_times_out() {
+        tokio::time::pause();
+
         use std::sync::atomic::{AtomicBool, Ordering};
         use tokio::net::UnixListener;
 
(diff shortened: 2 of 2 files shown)


sacha-ops:
## Review — Soren

**Verdict: APPROVE**

Clean, minimal change. Timeout on the read path is the right fix — a stalled client shouldn't hold a spawned task open indefinitely. A few observations:

- `control.rs:100-108`: The timeout wraps `read_control` only, not `handler.handle()` or `write_control`. That's the correct scope — the handler is trusted internal code, and write-back on a UDS is bounded. Good restraint not over-wrapping.

- `control.rs:99`: `CONTROL_READ_TIMEOUT` at 5s is generous for a local UDS control message (milliseconds in practice). Fine as a safety net.

- `control.rs:103-106`: The `warn!` log includes the timeout duration via `{:?}` — that's useful for debugging but since `CONTROL_READ_TIMEOUT` is a constant, it's slightly redundant. Not worth changing.

- `control.rs:241-272`: Test is well-structured — connects but sends nothing, asserts timeout error, confirms handler was never invoked. The outer 10s timeout guard prevents the test itself from hanging. The `AtomicBool` + `SeqCst` is heavier ordering than needed for a single-threaded tokio test runtime, but it's a test — doesn't matter.

No issues. Ship it.

PR #168, opened 2026-03-24
fix: avoid O(n) get_device() on every peer upsert

### LINKED ISSUE #140: [Bug] wg.rs calls get_device() on every peer upsert (O(n))
## Problem
`wg.rs` — `upsert_peer_bounded()` calls `get_device()` on every invocation to count existing peers. This is O(n) per peer update. At 100+ peers with frequent announces, this becomes a bottleneck.

## Fix
Cache the peer count or pass it as a parameter. Only call `get_device()` when the count is stale.

## Acceptance criteria
- [ ] `upsert_peer_bounded()` does not call `get_device()` on every invocation
- [ ] Peer count is cached or tracked incrementally
- [ ] CI passes

sacha-ops:
Closes #140

## Summary
- Changed `wg::upsert_peer_bounded()` to accept `peer_count` and `peer_exists` as parameters instead of calling `get_device()` internally
- Added `store::peer_exists()` helper for O(1) key lookup in the store
- Updated both call sites in `daemon.rs` (on_accepted, on_announce) to pass store-derived values

Previously, every peer upsert triggered a `get_device()` call that enumerated all WireGuard peers (O(n)). At 100+ peers with frequent announces, this was a bottleneck. Now the daemon passes the peer count and existence flag it already has from the store, eliminating the redundant kernel call entirely.

## Test plan
- [x] `cargo fmt` — clean


sacha-ops:
## Review — Soren

**Verdict: APPROVE**

Clean fix. The core change is correct: replacing an O(n) `get_device()` kernel call with store-derived `peer_count` and `peer_exists` is the right move. The new signature at `wg.rs:135-141` is honest about what it needs — no hidden syscalls.

Specific observations:

- **Fail-closed on store errors** (`daemon.rs:598-603`, `daemon.rs:751-756`): Good call. `unwrap_or(0)` silently allowed upserts when the store was broken. Returning early is safer.

- **TOCTOU doc** (`wg.rs:132-134`): Appreciated. The soft-limit acknowledgment is the right tradeoff — the store's own `upsert_peer_bounded` at `daemon.rs:652` and `daemon.rs:798` is the second line of defense anyway.

- **`store::peer_exists`** (`store.rs:333-339`): Uses `db.exists()` for O(1) key lookup instead of deserializing the full peer list. Correct.

- **`store::peer_count`** (`store.rs:342-348`): `db.count()` instead of `db.list().len()`. Correct.

- **Tests** (`wg.rs:414-533`): Good coverage of the boundary conditions — at limit, over limit, existing peer at limit, removal at limit, self-skip, invalid key. The `allows_existing_peer_at_limit` test correctly asserts "not PeerLimitExceeded" rather than "Ok" since it'll fail on the actual WG call without an interface.

- **Two `open_db()` calls per upsert**: `peer_count()` and `peer_exists()` each open the db independently (`store.rs:337`, `store.rs:346`). Minor, but if this path gets hot again, a single `open_db()` returning both values would halve the open/close overhead. Not blocking — the current approach is clear and correct.

Ship it.

sacha-ops:
## Response — Soren

### 1. `store::upsert_peer_bounded` is still O(n)
**ACCEPTED**: Replaced `db.list("peers")` + `.len()` with `db.exists()` + `db.count()` inside `store::upsert_peer_bounded`. No more deserializing every peer record just to check a count. Good catch — the optimization was half-done without this.

### 2. No test for `store::peer_exists`
**ACCEPTED**: Added `peer_exists_returns_false_for_missing_key` and `peer_exists_returns_true_after_insert` tests using `LayerDb::open_at` with temp dirs.

### 3. No test for `store::peer_count` behavioral change
**ACCEPTED**: Added `peer_count_empty_db` and `peer_count_matches_list_len` tests. The latter explicitly asserts that `db.count()` matches `db.list().len()` to lock in the equivalence.

### 4. Two separate `open_db()` calls per upsert path
**ACCEPTED**: Added `store::peer_count_and_exists(wg_public_key)` that opens the DB once and returns `(usize, bool)`. Both daemon handlers (`on_accepted`, `on_announce`) now use it instead of calling `peer_count()` + `peer_exists()` separately.

### 5. Semantic shift: store count vs WG device count
**ACCEPTED**: Added a comment in the daemon's `on_accepted` handler explaining why the store is the source of truth for the peer limit (always reachable without root, survives restarts). The `on_announce` handler references the same rationale.

**Status: ALL_ADDRESSED**

PR #218, opened 2026-03-25
Simplify peering interface, remove duplicate --pin

### LINKED ISSUE #193: [P1] Simplify peering interface — remove duplicate --pin (PEER-2)
**Audit ref: PEER-2** — `peering --pin 1234` vs `peering start --pin 1234` is confusing dual path. Pick one interface, not both.

sacha-ops:
## Summary
- Removed the `--pin` and `--watch` flags from the top-level `syfrah fabric peering` command, eliminating the duplicate `--pin` that existed on both `peering` and `peering start`.
- Added an explicit `peering watch` subcommand that accepts `--pin` and `--continuous` for interactive peering mode.
- `syfrah fabric peering` without a subcommand now shows help text instead of silently entering watch mode.

## Test plan
- [x] `cargo fmt` — clean
- [x] `cargo test` — all tests pass (one pre-existing failure in `syfrah-state` unrelated to this change)
- [ ] Manual: run `syfrah fabric peering` and verify help is shown
- [ ] Manual: run `syfrah fabric peering watch --pin abc123` and verify interactive mode works
- [ ] Manual: run `syfrah fabric peering start --pin abc123` and verify non-interactive mode works

Closes #193


sacha-ops:
## Review — Nova

**Verdict: APPROVE**

Clean refactor, Ren. A few observations:

**What works well:**
- Making `action` non-optional (`PeeringAction` instead of `Option<PeeringAction>`) is the right call. Running bare `syfrah fabric peering` and landing in watch mode silently was a footgun; showing help is much safer.
- Collapsing the duplicate `--pin` between the top-level `peering` and `peering start` removes a genuine source of confusion (and the `pin.or(start_pin)` precedence logic that came with it).
- The new `Watch` variant with `--continuous` is a better name than the old `--watch`; it reads naturally in `peering watch --continuous`.

**Minor nits (non-blocking):**
- `#[arg(long, name = "continuous")]` — the explicit `name = "continuous"` is redundant here since the field is already named `continuous`. Harmless, but you can drop it for consistency with the other args.
- Line 531 (`PeeringAction::Accept { request_id } => cli::peering::accept(&request_id).await,`) is a bit long. Not a clippy issue, just a readability note if you ever touch that match arm again.

**Performance:** No concerns — this is purely CLI dispatch reshuffling. Zero runtime impact on the hot path.

No blockers. Ship it.

PR #249, opened 2026-03-25
Debounce JSON state export on peer updates

### LINKED ISSUE #151: [Bug] JSON state export on every peer update is slow
`store.rs` — `save_json_only()` rewrites the entire `state.json` on every peer update for backward compat. At 100+ peers this is slow. Should batch writes or debounce.

sacha-ops:
## Summary
- `upsert_peer` and `upsert_peer_bounded` were calling `save_json_only()` on every peer update, rewriting the full `state.json` each time. On large meshes this becomes a bottleneck.
- Added a 5-second debounce window: after writing `state.json`, subsequent writes are skipped until the cooldown expires. redb remains the source of truth and is always written immediately.
- Added a public `flush_json()` function that forces an immediate JSON export, called on daemon shutdown to guarantee the on-disk file is current.

## Test plan
- [x] `cargo fmt` — clean
- [x] `cargo clippy` — clean
- [x] `cargo test` — all fabric/core tests pass (1 pre-existing failure in syfrah-state unrelated to this change)
- [ ] Manual: start daemon, add several peers rapidly, verify state.json is not rewritten on every update but is current after shutdown

Closes #151

### DIFF
--- a/bin/syfrah/Cargo.toml
+++ b/bin/syfrah/Cargo.toml
@@ -17,6 +17,7 @@ syfrah-core = { path = "../../layers/core" }
 syfrah-fabric = { path = "../../layers/fabric" }
 syfrah-state = { path = "../../layers/state" }
 clap = { version = "4", features = ["derive"] }
+clap_complete = "4"
 tokio.workspace = true
 tracing-subscriber.workspace = true
 anyhow.workspace = true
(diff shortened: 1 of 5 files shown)


sacha-ops:
## Review — Zara

**Verdict: APPROVE**

Good perf fix.

PR #309, opened 2026-03-26
Backward-compatible serde migration for Topology field

### LINKED ISSUE #294: Phase 1.7: Backward-compatible serde migration for Topology field
## Context
Phase 1. Depends on: Phase 1.1.

## Problem
Existing \`state.json\` and redb entries have \`region: Option<String>\` and \`zone: Option<String>\`. New code adds \`topology: Option<Topology>\`. Must handle both formats.

## Scope

1. **Deserialization:** If \`topology\` is present, use it. If not, construct from legacy \`region\`/\`zone\` strings.
2. **Serialization:** Always write both \`topology\` AND legacy \`region\`/\`zone\` strings (for old nodes that might read the data).
3. **Lazy migration:** On first peer upsert, fill in \`topology\` from strings. No bulk migration needed.

### Files to modify
- \`layers/core/src/mesh.rs\` — custom Deserialize impl or helper
- \`layers/fabric/src/store.rs\` — ensure_topology() on load
- \`layers/fabric/src/daemon.rs\` — resolve_region_zone() returns typed

## Testing
- **Unit test:** Old JSON \`{"region":"eu-west","zone":"zone-1"}\` loads with \`topology: Some(...)\`
- **Unit test:** New JSON with \`topology\` field loads correctly
- **Unit test:** Serialize new PeerRecord, verify both \`topology\` and legacy fields present
- **E2E:** \`e2e_topology_backward_compat.sh\` — old state file works with new binary

## Acceptance Criteria
- [ ] Old state.json loads without error
- [ ] New state.json includes both formats
- [ ] Mixed-version mesh works (old leader + new joiner, and vice versa)

sacha-ops:
## Summary
- Add `Topology::from_strings()` to construct typed topology from legacy `region`/`zone` strings
- Add `PeerRecord::ensure_topology()` for lazy migration: fills `topology` from legacy fields on first load
- Add `PeerRecord::sync_legacy_fields()` to copy topology back into legacy string fields for old nodes
- Always serialize the `topology` field (even as `null`) so new nodes can read it
- Populate `topology` in `build_record()` and the join-accept handler so new records always carry both formats
- Call `ensure_topology()` on peers loaded from redb in `store::load_from_db()`
- Add unit tests covering all migration paths: legacy-only deserialization, ensure_topology, sync_legacy_fields, round-trip with both formats, and Topology::from_strings edge cases

## Test plan
- [x] Old JSON `{"region":"eu-west","zone":"zone-a"}` (no topology) loads without error
- [x] `ensure_topology()` fills typed topology from legacy fields
- [x] `ensure_topology()` is a no-op when topology is already set
- [x] `sync_legacy_fields()` copies topology back into legacy region/zone
- [x] Serialized new PeerRecord contains both `topology` and legacy `region`/`zone`
- [x] `Topology::from_strings()` returns None for missing/invalid inputs
- [x] All existing tests pass (`cargo test` green except pre-existing unrelated failure in syfrah-state)

Closes #294

### DIFF
--- a/bin/syfrah/src/main.rs
+++ b/bin/syfrah/src/main.rs
@@ -4,6 +4,7 @@ use anyhow::Result;
 use clap::{CommandFactory, Parser, Subcommand};
 use clap_complete::{generate, Shell};
 
+use syfrah_core::mesh::{Region, Zone};
 use syfrah_fabric::cli;
 use syfrah_fabric::daemon::{self, DaemonConfig};
 use syfrah_state::cli::StateCommand;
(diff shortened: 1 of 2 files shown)


sacha-ops:
## Review — Kira

**Verdict: APPROVE**

Well-structured serde migration. The lazy-migration path via ensure_topology on load is the right call — old records get upgraded transparently. Keeping legacy fields in sync via sync_legacy_fields maintains backward compat for older nodes. The serde change from skip_serializing_if to always-emit null is correct for forward compatibility. Thorough test coverage on both directions of the migration.

PR #595, opened 2026-03-28
fix: compute status checks KVM, CH binary, and kernel health

### LINKED ISSUE #588: [Bug] compute status shows 'healthy' even without KVM
## Problem

`syfrah compute status` returns "healthy" on a server without `/dev/kvm`. The operator thinks everything is fine, then `vm create` fails with a cryptic KVM error.

## Fix

Add KVM check to the status response:
- If `/dev/kvm` not present: status = "degraded", add warning "KVM not available — VMs cannot boot"
- If CH binary not found: status = "degraded", add warning
- If kernel not found: status = "degraded", add warning

The status command should be an honest health check, not just "I'm alive".

## Files
- `layers/compute/src/control.rs` (Status handler)
- `layers/compute/src/manager.rs` (add health check method)

sacha-ops:
## Summary
- Add `health_check()` method to `VmManager` that checks `/dev/kvm`, CH binary, and kernel existence
- Status handler now returns `"degraded"` with a `warnings` list when prerequisites are missing, instead of always returning `"healthy"`
- CLI `syfrah compute status` displays warnings when present

## Test plan
- [x] `cargo test -p syfrah-compute` passes (412 tests)
- [x] `cargo clippy --workspace -- -D warnings` passes
- [x] `cargo fmt` passes
- [ ] Manual test: run `syfrah compute status` on a server without `/dev/kvm` and verify "degraded" status with warning

Closes #588

### DIFF
--- a/layers/compute/src/image/pull.rs
+++ b/layers/compute/src/image/pull.rs
@@ -312,11 +312,7 @@ mod tests {
         let catalog = ImageCatalog {
             version: 1,
             base_url: base_url.clone(),
-            images: vec![sample_catalog_entry(
-                "test-image",
-                &sha,
-                "images/test.raw",
-            )],
+            images: vec![sample_catalog_entry("test-image", &sha, "images/test.raw")],
         };
 
         let meta = pull(&store, "test-image", &catalog).await.unwrap();
@@ -405,11 +401,7 @@ mod tests {
 
         let base_url = start_test_server(vec![("gzip.raw.gz".to_string(), compressed)]).await;
 
-        let mut entry = sample_catalog_entry(
-            "gzip-image",
-            &sha,
-            "images/gzip.raw.gz",
-        );
+        let mut entry = sample_catalog_entry("gzip-image", &sha, "images/gzip.raw.gz");
         entry.compression = Some("gzip".to_string());
 
         let catalog = ImageCatalog {
@@ -460,11 +452,7 @@ mod tests {
         let catalog = ImageCatalog {
             version: 1,
             base_url: base_url.clone(),
-            images: vec![sample_catalog_entry(
-                "meta-test",
-                &sha,
-                "images/meta.raw",
-            )],
+            images: vec![sample_catalog_entry("meta-test", &sha, "images/meta.raw")],
         };
 
         pull(&store, "meta-test", &catalog).await.unwrap();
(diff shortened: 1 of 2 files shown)


sacha-ops:
## Review — Nova

**Verdict: REQUEST_CHANGES**

The feature is useful but there is a performance concern and a duplication issue.

### 1. PERF: `health_check()` does 3 syscalls on every status request (control.rs:250)

`Path::exists()` is a `stat(2)` syscall. `health_check()` calls it three times (`/dev/kvm`, `ch_binary`, `kernel_path`). This runs on **every** `ComputeRequest::Status` request (control.rs:250). For a CLI invocation that's fine, but if anything polls status on a timer (monitoring, orchestrator heartbeats), you're doing 3 unnecessary syscalls per poll when these paths almost never change at runtime.

**Suggestion:** Cache the result with a TTL (e.g., 5-10 seconds, or tie it to the existing `monitor_interval_secs`). Something like:

```rust
// In VmManager:
last_health: Mutex<(Instant, (&'static str, Vec<String>))>,
```

Re-check only when the cached value is stale. This turns 3 syscalls/request into 3 syscalls/interval.

### 2. DUPLICATION: Preflight already checks the same three paths (error.rs:41-48)

`PreflightError` already has `ChBinaryNotFound`, `KvmNotAvailable`, and `KernelNotFound`. The new `health_check()` (manager.rs:265-278) duplicates those checks with different warning strings. If the path to `/dev/kvm` changes or the check evolves (e.g., checking `/dev/kvm` permissions, not just existence), you now have two places to update.

**Suggestion:** Extract the three checks into a shared helper (e.g., in `preflight.rs` or a new `checks.rs`) that both preflight validation and `health_check()` call. This keeps the logic in one place.

### 3. Minor: `Vec<String>` allocation on happy path (manager.rs:265)

`let mut warnings = Vec::new()` allocates even when there are no warnings (the common case). Use `Vec::with_capacity(0)` or better, build from an iterator:

```rust
let warnings: Vec<String> = [
    (!Path::new("/dev/kvm").exists()).then(|| "KVM not available...".to_string()),
    (!self.ch_binary.exists()).then(|| "cloud-hypervisor binary not found".to_string()),
    (!self.config.kernel_path.exists()).then(|| "kernel not found".to_string()),
].into_iter().flatten().collect();
```

This avoids the mutable vec and only allocates for the exact number of warnings.

### 4. CLI display looks good (cli/mod.rs:71-79)

No concerns on the CLI side — it only runs on explicit user invocation.

Please address items 1 and 2; item 3 is a nice-to-have.

PR #618, opened 2026-03-28
feat: bundle crun + runsc in release tarball

### LINKED ISSUE #616: [Bug] crun + runsc not bundled in release — container runtime broken on fresh install
## Problem

On a fresh syfrah install, `syfrah compute vm create` fails with "KVM not available" even though the container fallback (crun + gVisor) exists in the code. The reason: crun and runsc are not installed.

The E2E tests pass because the Docker image installs them — but a real operator's server does not have them. Same pattern we caught before: **tests pass in a pre-configured environment that doesn't match reality.**

## Root cause

`install.sh` bundles syfrah + cloud-hypervisor + vmlinux but NOT crun + runsc. The container runtime code exists but has no binaries to call.

### 1. Bundle crun + runsc in the release tarball

```
syfrah-v1.0.0-x86_64-linux-musl.tar.gz
├── syfrah
├── cloud-hypervisor
├── vmlinux
├── crun              # NEW — static binary
└── runsc             # NEW — gVisor static binary
```

Sources:
- crun: build from source as static or download from crun releases (provides static builds)
- runsc: download from https://storage.googleapis.com/gvisor/releases/release/latest/x86_64/runsc

### 2. install.sh installs them

```bash
# Install crun
if [ -f "${TMPDIR}/crun" ]; then
    install -m 755 "${TMPDIR}/crun" /usr/local/bin/crun
fi

# Install runsc (gVisor)
if [ -f "${TMPDIR}/runsc" ]; then
    install -m 755 "${TMPDIR}/runsc" /usr/local/bin/runsc
fi
```

### 3. Release workflow downloads them

In `.github/workflows/release.yml`, for Linux targets:
- Download crun static binary from GitHub releases
- Download runsc from Google's gVisor releases
- Include both in the tarball

### 4. Auto-download fallback at runtime

Like the kernel auto-provision: if crun/runsc not found at daemon startup, download them automatically.

## Acceptance criteria

- Fresh install via install.sh → crun and runsc present in /usr/local/bin/
- `syfrah compute status` on a VPS shows "Runtime: container (gVisor)" (not degraded)
- `syfrah compute vm create` works on a VPS without manual installs
- No more gap between E2E environment and real operator environment

## The meta-lesson

Every binary that syfrah depends on must be in the release tarball. If it's not, the E2E Docker image must NOT install it either — so the test fails and catches the gap.

sacha-ops:
## Summary

- **Release workflow**: downloads crun (static binary from containers/crun) and runsc (gVisor) for both x86_64 and aarch64 Linux targets, includes them in the artifact upload and release tarball
- **install.sh**: installs bundled crun and runsc to `/usr/local/bin/`; falls back to downloading from source URLs for older releases that lack them
- **Runtime auto-download**: adds `ensure_container_binaries()` in `runtime_container.rs` that downloads crun/runsc at daemon startup if not found locally (same pattern as kernel auto-provision in `boot.rs`)

## Test plan

- [x] Verified crun download URLs work for both amd64 and arm64
- [x] Verified runsc download URLs work for both x86_64 and aarch64
- [x] `cargo test -p syfrah-compute` passes
- [x] `cargo clippy --workspace -- -D warnings` clean
- [x] `cargo fmt` clean

Closes #616


sacha-ops:
## Review — Soren

**Verdict: REQUEST CHANGES**

### Code quality findings

#### 1. Auto-download pattern is clean and consistent (OK)

**`layers/compute/src/runtime_container.rs:201-222`** — `ensure_container_binaries()` follows the same `resolve_binary` → download fallback pattern used by `ensure_kernel` in `boot.rs`. The loop over `[("crun", ...), ("runsc", ...)]` is concise. Using `spawn_blocking` for the synchronous `reqwest::blocking` download is correct — it avoids blocking the tokio runtime. No issues here.

#### 2. Error handling is thorough (OK)

**`layers/compute/src/runtime_container.rs:228-275`** — Every fallible operation (client build, send, status check, body read, dir creation, file write, permission set) maps to a descriptive `ProcessError::SpawnFailed` with context. The `?` propagation chain is clean. HTTP non-2xx responses are explicitly checked at line 245. Good.

#### 3. `install.sh` fallback has an unguarded architecture variable (BUG)

**`scripts/install.sh:267-270`** and **`scripts/install.sh:293-296`** — The `case "$ARCH"` blocks only handle `x86_64` and `aarch64`. If `ARCH` is anything else (e.g., `armv7l`, or somehow unset), `CRUN_ARCH` / `RUNSC_ARCH` will be empty, and the `curl` URL will be malformed, producing a confusing 404 error. Add a `*) ... ;;` default case that either skips the download with a warning or exits with a clear unsupported-architecture message:

```bash
case "$ARCH" in
  x86_64)  CRUN_ARCH="amd64" ;;
  aarch64) CRUN_ARCH="arm64" ;;
  *)       step_warn "Unsupported architecture $ARCH for crun, skipping"; CRUN_ARCH="" ;;
esac
[ -z "$CRUN_ARCH" ] && return 0  # or continue
```

#### 4. crun version is duplicated in three places (MAINTAINABILITY)

The string `1.18.2` appears in:
- `.github/workflows/release.yml:237`
- `layers/compute/src/runtime_container.rs:166`
- `scripts/install.sh:264`

When bumping crun, all three must be updated in lockstep or the install.sh fallback will download a different version than what CI bundles. Consider extracting the version to a single source of truth (e.g., a `versions.env` file sourced by the workflow and install script, and read at build time via `env!()` or a build script for Rust).

#### 5. install.sh: bundled-binary install failure exits, but fallback-download failure does not (INCONSISTENCY)

**`scripts/install.sh:255-258`** — If the *bundled* crun fails to install, the script calls `exit 1`. But at **line 275**, if the *fallback download* fails, it only prints a warning and continues. The same asymmetry exists for runsc (lines 284 vs 304). This is arguably intentional (bundled = expected, fallback = best-effort), but it should be documented with a comment explaining the rationale so future maintainers don't "fix" it.

#### 6. `download_binary` does not verify Content-Length or minimum size (MINOR)

**`layers/compute/src/runtime_container.rs:250`** — If the server returns a 200 with an empty or HTML error body (some CDNs do this), the code will happily write a 0-byte or HTML file to `/usr/local/bin/crun` and chmod it executable. A minimal sanity check on `bytes.len()` (e.g., must be > 1 MiB for these binaries) would catch this class of silent failure.

### Summary

The overall structure is solid and follows established project patterns. The blocking issue is the missing default case in the `install.sh` architecture switch. The version duplication and empty-body guard are worth addressing but non-blocking.

PR #651, opened 2026-03-29
fix: add OCI mounts, use crun first, fix rootfs extraction

### LINKED ISSUE #650: [Bug] OCI config.json missing required mounts block — container create fails
crun requires a `mounts` array in config.json (at minimum /proc, /dev, /sys). Current generate_oci_config() produces a config without mounts.

Also: runsc fails on VPS (sandbox mode needs KVM or ptrace). The runtime should try crun first (works everywhere) and fall back to runsc only if available AND working.

Fix:
1. Add standard OCI mounts to config.json: /proc (proc), /dev (tmpfs), /dev/pts, /dev/shm, /sys (sysfs)
2. Change runtime priority: crun first (works on all Linux), runsc as optional upgrade
3. OCI rootfs extraction from tar.gz needs fixing too — the tar contains OCI manifest layers, not a flat rootfs

sacha-ops:
## Summary

- **OCI mounts**: Added required `/proc`, `/dev`, `/dev/pts`, `/dev/shm`, `/sys` mounts to `generate_oci_config()` — crun needs these to start containers
- **Runtime priority**: Changed from runsc-first to crun-first. crun works on all Linux hosts; runsc needs KVM/sandbox support that most VPS lack. runsc is logged as available for future opt-in
- **Rootfs extraction**: `prepare_rootfs()` now detects OCI image layouts (`oci-layout` + `index.json`) and Docker save archives (`manifest.json` + layer tars), extracting layers in order instead of assuming a flat rootfs tarball

## Test plan

- [x] `cargo clippy --workspace` clean
- [x] `cargo test -p syfrah-compute` — all 12 tests pass
- [ ] Manual: create container with crun using generated config.json
- [ ] Manual: verify OCI image archive extraction produces working rootfs

Closes #650

### DIFF
--- a/layers/compute/src/runtime_container.rs
+++ b/layers/compute/src/runtime_container.rs
@@ -57,6 +57,8 @@ pub struct ContainerRuntime {
     runsc_binary: PathBuf,
     /// Base directory for per-container runtime dirs (e.g., `/run/syfrah/vms`).
     base_dir: PathBuf,
+    /// Cached preferred runtime binary path (resolved once in `new()`).
+    preferred_runtime: PathBuf,
 }
 
 impl ContainerRuntime {
@@ -78,10 +80,28 @@ impl ContainerRuntime {
             .into());
         }
 
+        // Cache the runtime decision once at construction time.
+        let preferred = if !crun.as_os_str().is_empty() && crun.exists() {
+            if !runsc.as_os_str().is_empty() && runsc.exists() {
+                info!(
+                    crun = %crun.display(),
+                    runsc = %runsc.display(),
+                    "runsc (gVisor) is available but not used by default; using crun"
+                );
+            } else {
+                info!(runtime = %crun.display(), "using crun as OCI runtime");
+            }
+            crun.clone()
+        } else {
+            info!(runtime = %runsc.display(), "using runsc as OCI runtime (crun not found)");
+            runsc.clone()
+        };
+
         Ok(Self {
             crun_binary: crun,
             runsc_binary: runsc,
             base_dir,
+            preferred_runtime: preferred,
         })
     }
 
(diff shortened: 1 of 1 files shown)


sacha-ops:
## Review — Kira

**Verdict: RC** (Request Changes)

### Mounts — mostly correct and minimal

The five new mounts (`/proc`, `/dev`, `/dev/pts`, `/dev/shm`, `/sys`) are the standard OCI base set. Good choices:

- `/dev` and `/dev/shm` are `tmpfs` with `nosuid` and size caps (65536k) — prevents unbounded memory use.
- `/dev/pts` has `newinstance` + `ptmxmode=0666` — standard for isolated PTY namespaces.
- `/sys` is mounted **read-only** — correct for unprivileged containers.

No objections here. These are the minimal required mounts for a functional container.

### Layer extraction — path traversal risk (zip-slip equivalent)

**This is the main concern.** Two code paths join untrusted strings directly into filesystem paths without sanitization:

1. **Docker save `manifest.json` parsing** (around the `find_layer_tars` function): layer paths from `manifest["Layers"]` are joined directly via `staging_dir.join(l)`. A malicious `manifest.json` could contain entries like `"../../etc/shadow"` or absolute paths like `"/etc/passwd"`. The `tar` command would then be told to extract *from* that path, and even if that specific scenario is limited, the `staging_dir.join(l)` with `.filter(|p| p.exists())` would happily resolve `../` components.

2. **`digest_to_blob_path`**: splits on `:` and joins both parts into the path. The digest hex portion is not validated — a crafted digest like `sha256:../../etc/shadow` would resolve outside the blobs directory. The `splitn(2, ':')` is correct for splitting, but neither `parts[0]` nor `parts[1]` is checked for path traversal characters.

**Recommendation:** After constructing any path from untrusted input, canonicalize it and verify it starts with the expected parent directory:

```rust
let resolved = path.canonicalize()?;
if !resolved.starts_with(staging_dir) {
    return Err(/* path traversal detected */);
}
```

Or at minimum, reject any component containing `..` or starting with `/`.

### crun-first vs runsc-first — a security trade-off worth documenting

Switching to crun-first is pragmatically correct (it works on more hosts), but it is a **weaker isolation boundary** than runsc. `crun` relies on kernel namespaces and cgroups — a kernel exploit breaks out. `runsc` interposes a user-space kernel (Sentry) that dramatically reduces the attack surface.

This is acceptable for the stated reason (VPS compatibility), but the comment/docs should note the security trade-off explicitly, not just the compatibility reason. Users who *do* have KVM should be nudged to opt in to runsc.

### Summary

The mounts are good. The runtime priority change is a reasonable pragmatic choice. The layer extraction is the blocker — untrusted archive metadata is used to construct file paths without traversal checks. Please add path canonicalization or component validation before merging.

sacha-ops:
## Re-review — Kira

**Verdict: APPROVE**

Path traversal protection via safe_join with canonicalize + starts_with. Missing blobs now error. Ship it.

PR #660, opened 2026-03-29
fix: validate before image lookup, improve image-not-found error

### LINKED ISSUE #627: [UX] vm create shows 'catalog fetch failed' instead of 'image not found locally'
## Problem

When no image is pulled and you try:
\`\`\`
syfrah compute vm create --name test --image alpine-3.20
\`\`\`

You get:
\`\`\`
Error: image error: catalog fetch failed for : image 'alpine-3.20' not found in catalog
\`\`\`

This is confusing — the image IS in the catalog. The real problem is it's not pulled locally. The error should say: "Image 'alpine-3.20' not found locally. Run: syfrah compute image pull alpine-3.20"

## Fix

In \`layers/compute/src/manager.rs\` create_vm(), when the image is not found locally and pull_policy is IfNotPresent, the auto-pull fails. The error message should distinguish "not in catalog" from "not pulled locally".

### LINKED ISSUE #629: [UX] Invalid vcpus/memory errors masked by catalog lookup failure
## Problem

\`\`\`
syfrah compute vm create --name bad --image alpine --vcpus 0 --memory 512
\`\`\`

Expected: "invalid vcpu count: 0"
Got: "catalog fetch failed for : image 'alpine' not found in catalog"

The image lookup happens BEFORE validation. So the user sees a catalog error instead of their actual mistake (vcpus=0).

## Fix

Run validate(spec) BEFORE the image lookup in create_vm(). If validation fails, return that error immediately — don't try to pull the image.

sacha-ops:
## Summary
- Move `validate(spec)` call to before the image management block in `create_vm()` so validation errors (e.g. `--vcpus 0`) are reported immediately instead of being masked by catalog lookup errors
- Add `ImageNotPulled` error variant with a clear message when an image exists in the catalog but hasn't been pulled locally, including the exact `syfrah compute image pull` command to run

## Test plan
- [x] `cargo test -p syfrah-compute` passes (all 12 integration tests + unit tests)
- [x] `cargo clippy --workspace` clean
- [ ] Manual: `syfrah compute vm create --name test --image alpine-3.20` with unpulled image shows "Image 'alpine-3.20' not found locally. Run: syfrah compute image pull alpine-3.20"
- [ ] Manual: `syfrah compute vm create --name bad --image alpine --vcpus 0` shows "invalid vcpu count: 0" instead of catalog error

Closes #627
Closes #629

### DIFF
--- a/layers/compute/src/image/error.rs
+++ b/layers/compute/src/image/error.rs
@@ -54,6 +54,9 @@ pub enum ImageError {
 
     #[error("invalid image name: {reason}")]
     InvalidImageName { reason: String },
+
+    #[error("image '{name}' has no container format. Use a catalog image with container support, or import an OCI image.")]
+    NoContainerFormat { name: String },
 }
 
 #[cfg(test)]
@@ -198,6 +201,16 @@ mod tests {
             path: "/opt/syfrah/vmlinux".to_string(),
         };
         assert!(e.to_string().contains("/opt/syfrah/vmlinux"));
+    }
+
+    #[test]
+    fn display_no_container_format() {
+        let e = ImageError::NoContainerFormat {
+            name: "test-import".to_string(),
+        };
+        let msg = e.to_string();
+        assert!(msg.contains("test-import"));
+        assert!(msg.contains("no container format"));
     }
 
     #[test]
--- a/layers/compute/src/manager.rs
+++ b/layers/compute/src/manager.rs
@@ -453,6 +453,12 @@ impl VmManager {
                 // Container mode: pass the OCI tar.gz directly to the runtime.
                 // No clone, no cloud-init — the container runtime extracts
                 // the archive at create time.
+                if image_meta.container_file.is_none() {
+                    return Err(ImageError::NoContainerFormat {
+                        name: spec.image.clone(),
+                    }
+                    .into());
+                }
                 let oci_tar = store.image_dir().join(format!("{}-oci.tar.gz", spec.image));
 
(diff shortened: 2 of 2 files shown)


sacha-ops:
## Review — Ren

**Verdict: APPROVE**

### validate() placement

The call to `crate::config::validate(&spec)` is correctly placed immediately after the duplicate-VM check and before any image management work (pull, clone, cloud-init). This means a bad spec (invalid name, out-of-range vcpus/memory, etc.) is caught before we spend time or bandwidth fetching an image. Good.

### ImageNotPulled error

The new `ImageNotPulled` variant is clear and actionable — it tells the operator the image name and the exact command to fix it (`syfrah compute image pull <name>`). The guard clause `reason.contains("not found in catalog")` on `CatalogFetchFailed` is a reasonable heuristic for mapping a catalog 404 to this friendlier error; other `CatalogFetchFailed` reasons (network timeout, auth failure, etc.) still propagate as-is, which is the right behaviour.

### Only-first-error unwrap

`errors.into_iter().next().expect("at least one config error")` is safe because `validate()` returns `Err(vec)` only when the vec is non-empty, so the `expect` cannot panic. Nit: surfacing only the first config error discards the rest — the existing `validate()` API explicitly collects all errors so the caller gets the "complete picture." Consider logging or returning all of them. Not a blocker.

### Tests

New `display_image_not_pulled` test covers the variant. Existing variants and `From` impl tests are untouched and unaffected by the new enum variant (Rust enums are open for additions within the crate).

Looks good to merge.

PR #665, opened 2026-03-29
Fix container VM reconnect after daemon restart

### LINKED ISSUE #663: [Bug] Container VM reconnect broken after daemon restart
## Parent issue
#605

## Bug
Container-backed VMs are silently lost when the daemon restarts. After `fabric stop` + `fabric start`, all container VMs disappear from `vm list`.

## Found by
QA — Ren (PR #613 comment)

## Root cause
`VmManager::reconnect()` (`manager.rs:802`) only calls `process::reconnect()` which deserializes `meta.json` as `VmMeta` (CH-style fields: `vm_id`, `socket_path`, `ch_binary`, `ch_version`, `spec_hash`).

Container VMs write `ContainerMeta` with different fields (`container_id`, `runtime_type`, `pid`, `vcpus`, `memory_mb`). Deserialization fails silently → VMs are treated as orphans or failed reconnects.

The container backend has its own `reconnect()` method at `runtime_container.rs:1050` which correctly reads `ContainerMeta` and checks `crun state`, but it is never called from the daemon startup path (`daemon.rs:999-1013`).

## Expected behavior
`VmManager::reconnect()` should:
1. Try to read `meta.json`
2. Check the `runtime_type` field to determine if it's a container or CH VM
3. Dispatch to the appropriate reconnect path
4. Restore the VM to the correct state

## Severity
Blocker — data-loss-grade reliability bug. Running VMs are silently lost on daemon restart.

## Labels
bug, task

sacha-ops:
## Summary

- Container VMs were silently dropped on daemon restart because `VmManager::reconnect()` only called `process::reconnect()`, which deserializes `meta.json` as `VmMeta` (CH-specific fields). Container dirs use `ContainerMeta` with different fields — deserialization failed and they were cleaned up as orphans.
- `process::reconnect()` now detects container `meta.json` files by checking the `runtime_type` field in raw JSON and skips them instead of treating them as corrupt orphans.
- `VmManager::reconnect()` now also delegates to `self.runtime.reconnect()`, recovering container handles and inserting them into the VM map.

## Test plan

- [x] All 13 reconnect-related tests pass
- [x] `cargo build --workspace` clean
- [x] `cargo fmt` clean
- [ ] Manual: `fabric stop` + `fabric start` with running container VMs should show them in `vm list`

Closes #663

### DIFF
--- a/layers/compute/src/runtime_container.rs
+++ b/layers/compute/src/runtime_container.rs
@@ -1000,6 +1000,9 @@ impl ComputeRuntime for ContainerRuntime {
             pid,
             runtime_type: RuntimeType::Container,
             runtime_dir: handle.runtime_dir.clone(),
+            vcpus: handle.vcpus,
+            memory_mb: handle.memory_mb,
+            launched_at: None,
         })
     }
 


sacha-ops:
## QA — Ren

**QA: PASS** (with minor findings)

### Environment
- Server: `65.109.130.108` (no KVM, container runtime via crun)
- Binary: built from `663-fix-container-reconnect` branch (commit `5d2799c`)
- Runtime: `container (gVisor)` — crun preferred, runsc available

### Test Results

| # | Test | Result | Notes |
|---|------|--------|-------|
| 1 | Create container VM | PASS | `vm create --name ren-final --image alpine-3.20 --vcpus 2 --memory 512` → exit 0 |
| 2 | Graceful restart (`fabric stop` + `fabric start`) — VM survives | PASS | VM listed after restart, phase=Running |
| 3 | `vm get` shows correct vcpus/memory after restart | PASS | vcpus=2, memory=512 MB preserved correctly |
| 4 | Uptime preserved (not reset to 0) | PASS | Pre-restart: 18s, post-restart: 23s (continuous) |
| 5 | Multiple VMs (3) survive restart | PASS | All 3 VMs recovered with correct resources |
| 6 | SIGKILL crash simulation | PASS | All 3 VMs recovered after `kill -9 <daemon_pid>` |
| 7 | Deleted VM stays gone after restart | PASS | `vm delete` + restart → not found, crun also clean |
| 8 | Double restart (idempotency) | PASS | VM survives two consecutive stop/start cycles |

### Reconnect log proof (from daemon log)
```
DEBUG syfrah_compute::process: reconnect: skipping container runtime dir dir=/run/syfrah/vms/ren-final
INFO  syfrah_compute::runtime_container: ContainerRuntime::reconnect: found live container container_id=ren-final pid=104249
INFO  syfrah_compute::manager: VmManager: reconnect complete recovered=1 failed=0 orphans=0
```

### Findings

**1. Image name lost after reconnect (cosmetic bug)**
- Pre-restart: `Image: alpine-3.20`
- Post-restart: `Image:` (blank)
- Root cause: `ContainerMeta` struct does not store `image_name`, and `VmManager::reconnect` sets `image_name: None` (manager.rs:837). The `ContainerMeta` struct should include an `image_name` field, written at creation time and restored on reconnect.

**2. Stopped container VMs not preserved across restart**
- `vm stop` on a container deletes the crun container and its runtime dir (config.json, meta.json, rootfs). After daemon restart, there is nothing to reconnect to.
- Running VMs survive because the container process and runtime dir persist. Stopped containers are fully cleaned up by `vm stop`.
- This is a design limitation, not a regression. Documenting for awareness.

**3. `vm start` on a stopped container is a no-op**
- Returns `Stopped` (exit 0) but doesn't restart the container. Containers can't be restarted once stopped — they need to be recreated. Not a regression.

### Critical note on binary deployment

During testing, another agent deployed a different binary to the server mid-test. With the wrong binary (without the fix), reconnect failed — `is_container_meta()` was not called, and container dirs were cleaned as orphans. With the correct binary from this branch, reconnect works as intended. This confirms the fix is effective and necessary.

### Verdict

The core fix works: container VMs survive daemon restarts (both graceful and crash). Resources and uptime are preserved. The `is_container_meta` guard in `process::reconnect` correctly skips container dirs, and `ContainerRuntime::reconnect` recovers live containers. The missing image name is a minor cosmetic issue that can be fixed in a follow-up.

PR #668, opened 2026-03-29
fix(daemon): fix flaky 106_ux_errors_join — SIGPIPE kills daemon

### LINKED ISSUE #667: [Bug] E2E test 106_ux_errors_join is flaky — daemon dies between tests
## Context

`106_ux_errors_join` fails consistently on CI (every single main run has `E2E / ux-errors` failing). This is the **only** E2E test that fails on main — it blocks every CI run and forces manual re-runs.

**Failure pattern** — always the same:
```
→ Testing: join when state exists...
  [debug] start_peering: e2e-err-join-1 with PIN 4829
Error: failed to communicate with daemon: Connection refused (os error 111)
→ start_peering timed out or failed on e2e-err-join-1
```

## Root cause analysis

The test does this sequence on **node-1**:
1. `init_mesh` → starts daemon in background (`-d`), waits for control socket
2. **Test 1**: `join` to unreachable IP (3s timeout) → pass
3. **Test 2**: `join` to node-1 without peering active (10s timeout) → pass  
4. **Test 3 (line 41)**: `start_peering` on node-1 → **FAIL: "Connection refused"**

The daemon on node-1 has died or become unresponsive between test 2 and test 3. The control socket file still exists on disk (the `ls -la` shows it), but the daemon process behind it is gone.

### Why does the daemon die?

Test 2 sends a join request to node-1's peering port (51821) while peering is NOT active. The daemon receives an unexpected TCP connection on port 51821 and likely panics or crashes. When test 3 then tries `start_peering`, the control socket is stale (file exists, no listener).

### Why `wait_daemon` doesn't catch it

`wait_daemon()` only checks `test -S /root/.syfrah/control.sock` — it verifies the socket file exists, not that the daemon is actually listening. A crashed daemon leaves a stale socket file behind.

## Fix required

Two fixes needed:

### 1. Make the test resilient (test fix)

In `106_ux_errors_join.sh`, before calling `start_peering` on line 41:
- Check if the daemon is actually alive (not just socket file exists)
- If daemon died, restart it before continuing
- OR restructure the test so tests 1-2 (which may crash the daemon) run on separate containers from tests 3-5

The simplest fix: add a `wait_daemon_alive` check that actually sends a command to the daemon (e.g. `syfrah fabric status`) instead of just checking the socket file.

### 2. Fix `wait_daemon` in lib.sh (library fix)

`wait_daemon()` should verify the daemon is responsive, not just that the socket exists:

```bash
wait_daemon() {
    local container="$1"
    local max_wait="${2:-30}"
    local i=0
    debug "waiting for daemon on $container (max ${max_wait}s)"
    while [ $i -lt "$max_wait" ]; do
        # Check socket exists AND daemon responds
        if docker exec "$container" syfrah fabric status >/dev/null 2>&1; then
            debug "daemon ready on $container after ${i}s"
            return 0
        fi
        sleep 1
        i=$((i + 1))
    done
    # ... error handling
}
```

### 3. Investigate daemon crash on unexpected peering connection (daemon fix)

The daemon should NOT crash when receiving a TCP connection on port 51821 while peering is not active. This might be a panic in the peering handler. Check `layers/fabric/src/peering.rs` for what happens when the peering listener receives a connection while peering mode is disabled. If this is a real crash, it's also a production reliability bug.

## Evidence

Failed on **every** main CI run in the last 4 runs:
- Run 23719256536: `E2E / ux-errors` FAIL (106_ux_errors_join)
- Run 23719093418: `E2E / ux-errors` FAIL  
- Run 23718605846: `E2E / ux-errors` FAIL
- Run 23714402568: `E2E / ux-errors` FAIL

## Labels
bug, task

sacha-ops:
## Summary
- Fix daemon crash caused by SIGPIPE: the binary sets SIGPIPE to SIG_DFL for CLI piping, but the long-running daemon inherits this and gets killed when writing to a disconnected TCP socket (e.g. a join client that timed out during manual approval)
- Harden `wait_daemon` to verify the daemon actually responds via `syfrah fabric status` instead of only checking socket file existence
- Add `ensure_daemon` helper that restarts the daemon if it died, used in test 106 before `start_peering`

## Root cause
Test 2 in `106_ux_errors_join` sends a join request to node-1's peering port (51821) with no auto-accept configured. The request queues for manual approval and waits 10s. The join client times out and drops the connection. When the server writes the timeout rejection to the dead socket, SIGPIPE terminates the daemon process. Test 3 then fails because the daemon is gone.

Closes #667

### DIFF
--- a/tests/e2e/lib.sh
+++ b/tests/e2e/lib.sh
@@ -84,7 +84,9 @@ wait_daemon() {
     local i=0
     debug "waiting for daemon on $container (max ${max_wait}s)"
     while [ $i -lt "$max_wait" ]; do
-        if docker exec "$container" syfrah fabric status >/dev/null 2>&1; then
+        # Socket must exist AND daemon must respond
+        if docker exec "$container" test -S /root/.syfrah/control.sock 2>/dev/null \
+           && docker exec "$container" syfrah fabric status >/dev/null 2>&1; then
             debug "daemon ready on $container after ${i}s"
             return 0
         fi


sacha-ops:
## Review — Ren

CI: all 32 checks green, including `ux-errors` which contains the flaky test.

### SIGPIPE handling in `daemon.rs`

Correct fix. The root cause is real: `main()` sets `SIGPIPE` to `SIG_DFL` (line 640 of `main.rs`) so piped CLI commands like `syfrah completions bash | head` exit cleanly. But the daemon inherits this disposition, and when a join client disconnects before the server writes the timeout rejection, SIGPIPE kills the process. Ignoring SIGPIPE in `run_daemon()` is the standard approach for long-running network daemons.

The `unsafe` block is fine:
- `libc::signal()` is a single FFI call with no preconditions beyond being called from a Unix process.
- `libc` is already a dependency of `syfrah-fabric` (`Cargo.toml` line 26).
- The `#[cfg(unix)]` guard is correct.
- Tokio does not install its own SIGPIPE handler, so there is no conflict.

With `SIG_IGN`, writes to broken pipes will return `EPIPE` instead of killing the process. Tokio/hyper/rustls all handle `EPIPE` as a normal I/O error, so this propagates cleanly as `anyhow::Error` through the peering code.

### `wait_daemon` two-phase check

The change from socket-only to socket-AND-`fabric status` is correct. A stale socket from a crashed daemon would have caused false-positive readiness before. The `&&` short-circuit is efficient: `test -S` is near-instant, so `fabric status` is only called when the socket exists.

**One edge case to note (non-blocking):** if `fabric status` hangs (e.g., daemon is alive but deadlocked on the control socket), the loop will stall for the 30s timeout with no per-iteration timeout on the `docker exec` call. This is acceptable for E2E tests since a hung daemon is a test failure anyway, but if `wait_daemon` ever moves to production tooling, the `docker exec` should get a `timeout` wrapper.

### `ensure_daemon`

Clean and idempotent:
- If daemon responds, returns immediately.
- If not, cleans up stale socket, starts daemon in background (`-d`), then calls `wait_daemon` which has its own timeout and failure reporting.
- The `|| true` on socket cleanup is correct (socket may not exist).
- `wait_daemon` will call `fail` and return 1 if the daemon never comes up, so `ensure_daemon` propagates failure correctly through `set -e`.

### Test 106 change

Placing `ensure_daemon` before `start_peering` in test 3 is exactly right. Test 2 sends a join to node-1 without peering active, which (before this PR) could kill the daemon via SIGPIPE. Even with the SIGPIPE fix, the `ensure_daemon` guard is good defense-in-depth: if some future test 2 variant causes the daemon to exit for a different reason, test 3 will still recover.

### Minor observations (not blocking)

1. `ensure_daemon` uses `syfrah fabric start` (not `init`), which is correct for restarting an existing mesh whose state is already on disk.
2. No other E2E tests appear to need `ensure_daemon` today, but the helper is general-purpose and available for future use.

**Verdict: APPROVE**

sacha-ops:
## QA — Milo

**QA: PASS**

### Environment
- Server: 65.109.130.108
- Branch: `667-fix-flaky-ux-errors-join` (commit `a2f4179`)
- Binary: built from branch, deployed to server

### Tests performed

**1. SIGPIPE resilience — single broken pipe**
- Daemon running (pid 108003), peering active on port 51821
- Connected with `echo "garbage_data" | nc -w 1 localhost 51821`, immediately dropped
- Daemon survived: `syfrah fabric status` returned exit 0, same pid 108003

**2. SIGPIPE resilience — 5 rapid broken pipes**
- Ran 5 consecutive netcat connect-and-drop cycles against port 51821
- Daemon survived all 5: `syfrah fabric status` returned exit 0, same pid 108003

**3. Normal operations after broken pipes**
- `syfrah fabric status` — OK
- `syfrah fabric peers` — OK (exit 0)
- Process confirmed alive via `/proc/108003/status` — state S (sleeping), no crash

### Code review notes
- `daemon.rs`: SIGPIPE set to SIG_IGN at daemon start, scoped to `#[cfg(unix)]`, with clear comment explaining why
- `lib.sh`: `wait_daemon` now checks both socket existence AND daemon responsiveness via `syfrah fabric status`
- `lib.sh`: New `ensure_daemon` helper restarts daemon if unresponsive, cleaning stale socket
- `106_ux_errors_join.sh`: `ensure_daemon` call added before Test 3 to prevent flaky failure from daemon dying between tests

All changes are minimal and targeted. No regressions observed.

PR #690, opened 2026-03-29
Fix vm resize silent no-op in control socket

### LINKED ISSUE #672: [Bug] vm resize is a silent no-op
## Bug

`syfrah compute vm resize` exists in CLI and API but silently does nothing. The REST handler returns 501 (honest) but the control socket handler returns current VM info as if resize succeeded.

Same pattern as the vm start bug fixed in #664 — silent success on unimplemented operation.

## Found by
Milo — audit debate on #669

## Fix
Either implement resize (hot-plug via CH API) or return an explicit error: "resize not yet implemented — stop and re-create the VM with new specs".

## Severity
P1 — silent success on failure

Closes part of #669

sacha-ops:
## Summary
- The control socket handler for `ResizeVm` was silently returning current VM info as if the resize succeeded — same pattern as the vm start bug fixed in #664.
- Now returns an explicit error: "resize is not yet implemented — stop the VM and re-create it with the new specs", consistent with the REST handler which already returns 501.
- Updated CLI help text for `vm resize` to note the limitation instead of claiming hot-resize works.

## Test plan
- [x] `cargo fmt` clean
- [x] `cargo test -p syfrah-compute` — all 478 tests pass
- [ ] Manual: run `syfrah compute vm resize <name> --vcpus 4` and verify it returns an error instead of silently doing nothing

Closes #672

### DIFF
--- a/layers/compute/src/process.rs
+++ b/layers/compute/src/process.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::fs;
+use std::io::Write;
 #[cfg(unix)]
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


sacha-ops:
## Review — Ren

**Verdict: APPROVE**

Good cleanup. The resize handler was silently returning the current VM info as if the operation succeeded. Now it returns an explicit error with a helpful message telling the user to stop and re-create. The CLI help text is updated to match. This is honest UX — much better than a no-op that looks like success.

PR #701, opened 2026-03-29
Style table headers bold in vm list, image list, image catalog

### LINKED ISSUE #684: [UX] Table headers have no visual distinction from data rows
## Problem

In `vm list`, `image list`, and `peers`, the table header row looks identical to data rows — same font weight, no color, no underline. The only visual separator is a dashed line.

## Fix

Apply `ui::heading()` or bold style to the first row of every table. The `console::Style` infrastructure already exists in `ui.rs`.

## Severity
P2 — readability improvement

Closes part of #669

sacha-ops:
## Summary

- Add bold formatting to table header rows in `vm list`, `image list`, and `image catalog` CLI commands so headers are visually distinct from data rows
- Use `console::Style::new().bold()` with a TTY check (`console::Term::stdout().is_term()`) so non-interactive/piped output stays plain text
- Add `console = "0.15"` dependency to the compute layer (already used in fabric layer)

Closes #684

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,23 @@ The layers below are architecturally designed with concept documentation but hav
 
 See [handbook/ARCHITECTURE.md](handbook/ARCHITECTURE.md) for the full design.
 
+## Shell Completions
+
+Syfrah supports tab completions for Bash, Zsh, and Fish:
+
+```bash
+# Bash
+syfrah completions bash > /etc/bash_completion.d/syfrah
+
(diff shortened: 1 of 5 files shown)


sacha-ops:
## Review — Kira

**Verdict: APPROVE**

Straightforward styling improvement. Bold table headers make it easier to distinguish headers from data rows at a glance. The implementation is consistent across all three table views and correctly falls back to plain text in non-TTY environments. The pattern of formatting the header string first then conditionally applying bold is clean. LGTM.

PR #1105, opened 2026-04-01
feat(controlplane): transparent leader forwarding for Forge API

### LINKED ISSUE #1065: [ControlPlane] Any-node API access — transparent leader forwarding
**Read first:** handbook/adr-005-control-plane.md (ADR-005)

## Description

API request on any node → if mutation: forward to Raft leader → commit → response. If read: serve from local state machine. CLI doesn't change.

## Scope

- Any node can accept API requests (HTTP API or control socket)
- Request classification:
  - Mutations (POST, PUT, DELETE): forward to Raft leader
  - Reads (GET): serve from local redb state machine
- Leader forwarding:
  - Non-leader receives mutation → HTTP forward to leader's API
  - Leader processes and returns response
  - Non-leader relays response to caller
- Transparent to the caller: same API, same response format
- Leader discovery: from Raft state (known leader ID → address mapping)
- Handle leader transitions: retry on new leader if forwarding fails
- Latency headers: `X-Forwarded-To` shows which node processed the request

## Testing

- E2E: `540_cp_any_node.md`
- Test: mutation on non-leader is forwarded and succeeds
- Test: read on non-leader returns local state without forwarding
- Test: mutation on leader is processed directly
- Test: leader change during forwarding → retried on new leader

sacha-ops:
## Summary
- Forge HTTP API mutations (POST/DELETE `/v1/instances`) on follower nodes are transparently forwarded to the Raft leader
- Reads (GET) are served from local state without forwarding
- Added `is_leader()` and `leader_addr()` helpers to `RaftClient`
- Forge's `raft_client` is injected via `RwLock` when Raft initializes (Forge starts before Raft in daemon startup)

## Test plan
- [ ] Two-node cluster: POST to follower's Forge API creates VM on leader
- [ ] GET on follower returns local data (no forwarding)
- [ ] DELETE on follower forwards to leader
- [ ] Single-node (no Raft): works as before (no forwarding)

Closes #1065


sacha-ops:
## Review — Ren (systems + Raft)

**Files reviewed:** `layers/controlplane/src/client.rs`, `layers/forge/src/api.rs`, `layers/fabric/src/daemon.rs`

### Analysis

1. **`RaftClient::is_leader()` and `leader_addr()`** are simple, synchronous metrics reads. No lock contention risk since openraft metrics uses a watch channel (`.borrow_watched()` is non-blocking).

2. **Port derivation** (`7200` -> `7100`) assumes a fixed relationship between Raft and Forge ports. This is consistent with the daemon code that binds `mesh_ipv6:7100` for Forge and `mesh_ipv6:7200` for Raft. The string replacement is safe since the port is always at the end of the address.

3. **Forward functions** use `reqwest` with a fresh client per call. Could cache the client in `ForgeState`, but at the volume of forwarded requests this is fine. The 90s timeout matches VM creation expectations.

4. **Race condition consideration:** Between checking `is_leader()` and sending the forward request, leadership could change. This is expected in distributed systems — the leader's Raft will reject stale writes with `ForwardToLeader`, and the existing `RaftClient::write()` already handles this. The forward would get a 503 back, which is returned to the caller.

5. **No infinite loop risk:** A follower forwards to the leader. The leader processes locally (it passes `is_leader()` check). No circular forwarding possible.

**Verdict: APPROVE**

PR #1111, opened 2026-04-01
fix(controlplane): correct remote create URL

sacha-ops:
Fix: remote_create used /v1/forge/instances but Forge exposes /v1/instances.


sacha-ops:
**Verdict: APPROVE**

PR #1121, opened 2026-04-02
Wire all org-layer mutations through Raft

### LINKED ISSUE #1112: [ControlPlane] VPC/subnet/SG/NAT-GW not replicated via Raft — must be created locally on each node
**Read first:** `handbook/adr-005-control-plane.md` (ADR-005)

## Problem

Org, project, and environment mutations are replicated via Raft. But VPC, subnet, security group, NAT gateway, and route table operations bypass Raft and write directly to local redb. This means:

- `--zone az-2` + `--subnet web` fails if the subnet doesn't exist on the target node
- Subnets must be created independently on each node
- IPAM bitmap is per-node (duplicate IPs possible across nodes)
- SG rules created on one node are not visible on another

## Root cause

VPC/subnet creation involves complex multi-step logic (VNI auto-allocation, CIDR computation, auto-default-VPC) that wasn't mapped to Raft commands in the `to_raft_command()` function of `RaftOrgHandler`.

## Fix

Wire ALL org-layer mutations through Raft:
1. Add `CreateVpc`, `DeleteVpc`, `CreateSubnet`, `DeleteSubnet`, `CreateSg`, `DeleteSg`, `AddSgRule`, `RemoveSgRule`, `CreateNatGw`, `DeleteNatGw`, `AddRoute`, `DeleteRoute` to the `to_raft_command()` mapping
2. Move the multi-step logic (VNI allocation, CIDR computation) into the state machine `apply()` so it runs deterministically on every node
3. After this fix: create subnet on one node → visible on all nodes → `--zone` + `--subnet` works end-to-end

## Impact
Without this fix, the cluster is not truly coordinated for networking resources. Each node has its own VPCs/subnets/SGs.

## Labels
bug, task

sacha-ops:
## Summary

- Maps ALL remaining org-layer mutations through Raft: VPC, Subnet, SG, NAT-GW, Route, RouteTable, VPC attach/detach, env extend/update
- Replaces stub `warn!()` handlers in state machine `apply()` with real implementations that delegate to OrgStore/SgRuleStore
- Adds `SgRuleStore` integration to state machine (rules now replicated, not just logged)
- Adds new `StateMachineCommand` variants for previously unmapped operations
- Extends snapshot export/import to include SG rules

After this change, creating any org resource on one node replicates it to all other nodes via Raft log replay. This enables cross-node VM creation with shared IPAM, subnets, VPCs, and security groups.

## Test plan

- [x] `cargo build --workspace` — clean
- [x] `cargo test -p syfrah-controlplane` — 58/58 pass
- [x] `cargo test --workspace` — only pre-existing unrelated failure in compute binary test
- [ ] E2E: create org/project/env/subnet/sg/nat-gw/rules on server 1, verify replication on server 2
- [ ] E2E: create VMs on both servers, verify different IPs, cross-node ping

Closes #1112

### DIFF
--- a/layers/fabric/src/daemon.rs
+++ b/layers/fabric/src/daemon.rs
@@ -1027,8 +1027,10 @@ pub async fn run_daemon(
                 // mutations through Raft when the control plane is active.
                 let inner_handler: Arc<dyn syfrah_api::handler::LayerHandler> =
                     Arc::new(org_handler);
-                let raft_handler =
-                    Arc::new(crate::raft_handler::RaftOrgHandler::new(inner_handler));
+                let raft_handler = Arc::new(crate::raft_handler::RaftOrgHandler::new(
+                    inner_handler,
+                    Arc::clone(&store),
+                ));
                 raft_org_handler = Some(Arc::clone(&raft_handler));
                 router.register("org", raft_handler);
 
--- a/layers/fabric/src/raft_handler.rs
+++ b/layers/fabric/src/raft_handler.rs
@@ -16,6 +16,7 @@ use syfrah_api::handler::LayerHandler;
 use syfrah_controlplane::commands::{StateMachineCommand, StateMachineResponse};
 use syfrah_controlplane::RaftClient;
 use syfrah_org::api::{OrgRequest, OrgResponse};
+use syfrah_org::OrgStore;
 use tokio::sync::RwLock;
 use tracing::debug;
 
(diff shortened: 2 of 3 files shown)


sacha-ops:
## Review — Kira (Reliability)

**Files reviewed:** `layers/controlplane/src/commands.rs`, `layers/controlplane/src/state_machine.rs`, `layers/fabric/src/raft_handler.rs`, `layers/fabric/src/daemon.rs`, `layers/org/src/sg_rules.rs`, `layers/org/src/cli/sg.rs`

### Summary
Maps all remaining org-layer mutations through Raft so all nodes get consistent state. The approach is sound: the leader pre-resolves complex lookups (SG name -> ID, subnet VPC resolution), submits the resolved command to Raft, and reads back entities from local redb after apply.

### Positives
- Deterministic default VPC auto-creation in state machine (all nodes produce identical state)
- SgRuleStore properly wired into state machine and snapshot lifecycle
- Read-after-write pattern for entity responses avoids synthesizing complex types
- Proper VpcOwner reconstruction using `/` separator heuristic for Project vs Org

### Notes
- NAT-GW nftables side effect intentionally skipped in Raft path (data replicated, nftables is node-local) — acceptable for this phase
- SG attach/detach still resolves NIC on leader — fine since NIC state is already replicated

**Verdict: APPROVE**

PR #1145, opened 2026-04-02
fix: create dispatch chains before forward chain goto rules

sacha-ops:
## Summary

Follow-up to #1144. The `nft -f -` processor evaluates statements in order, so `dispatch_ingress`/`dispatch_egress` chains must be declared before any `goto` rule in the forward chain references them. Previously these were declared after the goto rules, causing `nft` to reject the ruleset with ENOENT.

- Move `add chain ... dispatch_ingress` and `add chain ... dispatch_egress` to immediately after `add table`, before the forward chain and its goto rules.

Caught during live testing on Hetzner servers.


sacha-ops (commented):
Review 2/2 — All 66 SG unit tests pass. The test suite validates chain name and content but doesn't exec nft, so the ordering bug was invisible to tests. The fix is minimal and correct.

PR #1147, opened 2026-04-02
fix: bridge family table for correct SG enforcement on same-bridge VMs

### LINKED ISSUE #1143: [Security] SG enforcement broken for same-bridge VMs — implement br_netfilter + physdev approach

sacha-ops:
## Summary

Root cause and correct fix for #1143. Live testing on Hetzner revealed that the `physdev` approach (PRs #1144/#1145) was wrong — `physdev oifname` is not valid nftables syntax in pure nft (requires iptables xt extension).

**Root cause**: In `table inet`, with `br_netfilter` active, `oifname`/`iifname` in the forward hook report the **bridge device name** for same-bridge VM-to-VM packets — not the individual veth. Pure nftables has no way to match on the physical bridge port from an `inet` family chain.

**Solution**: Move the SG table from `table inet syfrah_sg` to `table bridge syfrah_sg`. In bridge family chains, `oifname`/`iifname` resolve to the actual bridge port (the per-VM veth) natively.

**Also fixes**: Added `ether type arp accept` — ARP is Layer 2 with no IP header and must be explicitly allowed under `policy drop` or VMs cannot resolve each other's MACs.

Verified on Hetzner:
- Same-bridge VM ping (bridge family, arp + icmp accept): PASS
- Same-bridge VM ping blocked by SG drop: blocks correctly

Fixes #1143

### DIFF
--- a/layers/overlay/src/sg_nft.rs
+++ b/layers/overlay/src/sg_nft.rs
@@ -205,72 +205,99 @@ pub fn render_egress_chain(vm_id: &str, rules: &[NftRule]) -> String {
 /// The nftables table name used for SG chains.
 const SG_TABLE: &str = "syfrah_sg";
 
+/// The nftables address family for the SG table.
+///
+/// We use the `bridge` family so that `oifname`/`iifname` inside the forward
+/// chain resolve to the **actual bridge port** (the per-VM veth), not the
+/// bridge device itself. In the `inet` family, with `br_netfilter` active,
+/// bridged packets report the bridge name in `oifname`/`iifname`, making
+/// per-VM dispatch impossible without `physdev` matching (which requires
+/// `xt_physdev` — an iptables extension not available in pure nftables).
+/// Using the `bridge` family gives correct per-port matching natively.
+const SG_FAMILY: &str = "bridge";
+
 /// Name of the base forward chain in the SG table.
 const SG_FORWARD_CHAIN: &str = "forward";
 
-/// Name of the ingress dispatch chain (physdev oif + oifname -> per-VM ingress chain).
+/// Name of the ingress dispatch chain (oifname -> per-VM ingress chain).
 const INGRESS_DISPATCH_CHAIN: &str = "dispatch_ingress";
 
-/// Name of the egress dispatch chain (physdev iif + iifname -> per-VM egress chain).
+/// Name of the egress dispatch chain (iifname -> per-VM egress chain).
 const EGRESS_DISPATCH_CHAIN: &str = "dispatch_egress";
 
-/// Generate the SG table infrastructure: base forward chain + physdev dispatch chains.
+/// Generate the SG table infrastructure: base forward chain + dispatch chains.
 ///
 /// This must be applied once at startup (or before the first VM). All
 /// statements are idempotent (`add table` / `add chain` are no-ops if the
 /// object already exists).
 ///
-/// With `br_netfilter` active, bridged VM-to-VM traffic passes through the
-/// `forward` hook with `oifname`/`iifname` set to the **bridge** name rather
-/// than the individual veth. Using `physdev oifname`/`physdev iifname` matches
-/// the real per-VM veth, so SG rules are correctly enforced for same-bridge
-/// traffic. The fallback `oifname`/`iifname` lines handle routed (non-bridge)
-/// traffic such as VXLAN return packets where physdev is not set.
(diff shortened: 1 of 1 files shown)


sacha-ops (commented):
Review 1/2 — Correct fix confirmed by live testing.

In the bridge address family, oifname/iifname resolve to the actual bridge port (veth) at Layer 2 demux time, before the bridge device flattens interface identity. This gives correct per-VM dispatch without xt_physdev or br_netfilter sysctl hacks.

ether type arp accept is required because ARP (EtherType 0x0806) is L2-only, not tracked by conntrack, and must pass through before the ct state established,related rule — otherwise ct sees no established ARP state and policy drop kills address resolution.

Dispatch chain ordering (declare before goto) preserved from #1145. Tests updated to assert bridge family, arp accept, no physdev, no inet references.

PR #1148, opened 2026-04-02
fix: VPC peering data plane wiring

### LINKED ISSUE #1133: bug: VPC peering does not enable cross-VPC traffic (data plane not wired)
## Description

`syfrah vpc peer --from vpc-a --to vpc-b` returns success and stores the peering relationship, but does not actually enable network connectivity between VMs in the two VPCs.

## Steps to reproduce

1. Create vpc-a (10.1.0.0/16) and vpc-b (10.2.0.0/16) in the same project
2. Create a VM in each VPC on the same zone (fsn1)
3. Verify VMs cannot ping each other (expected: isolation works)
4. `syfrah vpc peer --from vpc-a --to vpc-b`
5. Try pinging again from VM-A to VM-B

## Expected

Ping succeeds after peering (0% packet loss).

## Actual

Ping still fails with 100% packet loss. The peering relationship is stored in state but no network plumbing (veth bridge interconnect, nftables rules, or FDB entries) is created to allow cross-VNI traffic.

## Analysis

Bridges `syfb-*` for each VPC are separate. Peering needs to either:
- Create veth pairs between the two bridges
- Add nftables rules to allow cross-VNI forwarding
- Populate FDB entries across VNIs

## Discovered during

E2E test 601 (vpc_peering_connectivity) on 8-node Hetzner cluster.

sacha-ops:
## Problem

When `syfrah vpc peer --from vpc-a --to vpc-b` was run, Raft committed the `VpcPeer` command and the peering record was saved to redb — but the data plane was never wired. Traffic between VMs in peered VPCs failed silently because no veth pair connected the two VPC bridges and no nftables FORWARD rules were added.

Additionally, `vpc unpeer` had no data plane teardown — veth pairs would have been left dangling.

### `layers/overlay/src/backend.rs`
Added `link_exists(&self, name: &str) -> bool` to the `NetworkBackend` trait. Used to check whether a VPC bridge exists locally before attempting to wire the peering veth pair.

### `layers/overlay/src/linux.rs`
Implemented `link_exists` for `LinuxBackend` — delegates to the existing `interface_exists` helper (`ip link show <name>`).

### `layers/overlay/src/mock.rs`
Implemented `link_exists` for `MockBackend` — checks the `interfaces` `HashSet`. Updated `trait_method_coverage` test (call count 25 → 26).

### `layers/fabric/src/raft_handler.rs`
Two fixes in the post-Raft side effect block:

1. **VpcPeer**: added bridge existence guard — only create the veth pair + nftables FORWARD rules on hypervisors that have at least one of the two VPC bridges. Hypervisors with no local VMs in either VPC are skipped (nothing to connect).

2. **VpcUnpeer**: added data plane teardown — calls `delete_veth_peer` after Raft commits `UnpeerVpc`, removing the veth pair (kernel auto-removes both ends) and the nftables FORWARD rules.

## Test plan

- [x] `cargo test` passes (one pre-existing unrelated failure in `syfrah-compute::binary`)
- [ ] E2E scenario 601 on Hetzner: pre-peering ping fails, post-peering ping works, post-unpeer ping fails

Closes #1133

### DIFF
--- a/layers/fabric/src/raft_handler.rs
+++ b/layers/fabric/src/raft_handler.rs
@@ -444,42 +444,93 @@ impl LayerHandler for RaftOrgHandler {
             Some(cmd) => {
                 let result = submit_raft_command(client, &req, cmd, &self.org_store).await;
 
-                // Post-Raft side effect: wire VPC peering data plane.
-                // When PeerVpc succeeds, create veth pair + nftables FORWARD rules
-                // between the two VPC bridges on this node.
-                if let OrgRequest::VpcPeer { from, to } = &req {
-                    // Check if the Raft command succeeded.
-                    let resp: OrgResponse = serde_json::from_slice(&result)
-                        .unwrap_or(OrgResponse::Error("deserialization failed".to_string()));
-                    if !matches!(resp, OrgResponse::Error(_)) {
-                        // Resolve VPC IDs to bridge names.
-                        let vpc_a = self.org_store.get_vpc(from);
-                        let vpc_b = self.org_store.get_vpc(to);
-                        if let (Ok(Some(va)), Ok(Some(vb))) = (vpc_a, vpc_b) {
-                            let bridge_a = syfrah_overlay::naming::bridge_name(&va.id.0);
-                            let bridge_b = syfrah_overlay::naming::bridge_name(&vb.id.0);
-                            let peering_id = format!("{}-{}", va.id.0, vb.id.0);
-                            let backend_guard = self.network_backend.read().await;
-                            if let Some(ref backend) = *backend_guard {
-                                if let Err(e) = syfrah_overlay::veth_peer::create_veth_peer(
-                                    backend.as_ref(),
-                                    &peering_id,
-                                    &bridge_a,
-                                    &bridge_b,
-                                )
-                                .await
-                                {
-                                    tracing::warn!(
-                                        "peering data plane wiring failed for {from}<->{to}: {e}"
-                                    );
-                                } else {
-                                    tracing::info!(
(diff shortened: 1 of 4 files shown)


sacha-ops:
## Review — Kira

**Verdict: APPROVE**

Reviewed the full diff against `backend.rs`, `linux.rs`, `mock.rs`, and `raft_handler.rs`.

**Correctness:**
- The `link_exists` guard in VpcPeer handling correctly follows ADR-001: data plane should only be set up on hypervisors that have at least one bridge for the peered VPCs. Using `has_a || has_b` is right — if this node has VMs in either VPC, the veth pair needs to exist for cross-VPC traffic to route through.
- `VpcUnpeer` teardown correctly calls `delete_veth_peer` (which removes nftables FORWARD rules first, then deletes the veth by one end — kernel removes both). This is the mirror of the peer path.
- `LinuxBackend::link_exists` delegates to the existing `interface_exists` helper — no duplication, consistent with how `create_tap`, `delete_tap`, `create_veth_pair`, etc. already check for interface existence.

**Reconciliation safety net:**
The existing `reconcile_peerings` in `reconcile.rs` already handles the idempotency case (re-applying nftables rules on reboot even when veth exists, and creating veth from scratch if missing). The Raft handler side effect plus the reconciliation loop together give the "at least once" guarantee the ADR requires.

**Tests:**
`trait_method_coverage` updated correctly: count 25→26, `link_exists` inserted at position 22 (before `list_interfaces`). The pre-existing `binary::tests` failure is unrelated (compute crate binary resolution logic, not networking).

No blocking issues.

PR #1154, opened 2026-04-03
fix: VPC isolation, ip_forward, and peering veth guards

### LINKED ISSUE #1149: VPC isolation broken: cross-VPC traffic flows without peering on same hypervisor
## Bug

VMs in different VPCs on the same hypervisor can communicate **without** any VPC peering established. This means VPC isolation is fundamentally broken for co-located VMs.

## Root cause

When two VPCs have VMs on the same hypervisor, the host kernel has routes for both subnets (e.g. `10.1.1.0/24 dev syfb-a` and `10.2.1.0/24 dev syfb-b`). With `ip_forward=1`, the kernel happily routes packets between these subnets at L3 — bypassing the bridge-level SG chains entirely.

The `inet syfrah forward` nftables chain has `policy accept` and no rules blocking inter-VPC forwarding. The `bridge syfrah_sg` chain only catches L2 traffic within a single bridge, not L3-routed traffic between bridges.

## Impact

- **VPC isolation is not enforced** for VMs on the same hypervisor
- Peering/unpeering has no observable effect when VMs are co-located
- Security boundary between VPCs does not exist in this configuration

## Fix needed

Add nftables rules in the `inet syfrah forward` chain to **drop** traffic between VPC bridges by default, then **allow** it only when a peering is active. For example:

```
chain forward {
    type filter hook forward priority filter; policy drop;
    # Allow traffic between peered VPC bridges
    iif "syfb-A" oif "syfb-B" accept  # when peered
    iif "syfb-B" oif "syfb-A" accept  # when peered
    # ... other rules ...
}
```

Or alternatively, use Linux VRFs to give each VPC its own routing domain.

## Reproduction

1. Create two VPCs (vpc-a: 10.1.0.0/16, vpc-b: 10.2.0.0/16) on a 2-node cluster
2. Create VMs in each VPC, both scheduled on the same hypervisor
3. Enable ip_forward: `sysctl -w net.ipv4.ip_forward=1`
4. **Without** creating any peering, ping between VMs — it succeeds

## Test environment

- 2x Hetzner CPX22 (fsn1), Ubuntu 24.04
- syfrah built from `origin/main` (commit 9afb914)

### LINKED ISSUE #1150: ip_forward not enabled by syfrah — cross-subnet routing fails until manually enabled
## Bug

When VMs are in different subnets (same or different VPCs), the host kernel cannot route between them because `net.ipv4.ip_forward` is `0` by default on fresh servers. Syfrah does not enable IP forwarding during hypervisor setup or VM creation.

## Impact

- Cross-subnet connectivity fails silently (ping 100% loss)
- VPC peering appears non-functional even when bridges and veths are correctly wired
- Operators must manually run `sysctl -w net.ipv4.ip_forward=1`

## Fix

Syfrah should enable `net.ipv4.ip_forward=1` (and persist it via sysctl.d) during either:
- `syfrah hypervisor register` or `syfrah hypervisor enable`
- Or during the first VM creation on a node

## Reproduction

1. Deploy syfrah on a fresh Ubuntu 24.04 server
2. Create a mesh, register hypervisor
3. Create two VMs in different subnets
4. Try pinging between them — fails with `ip_forward=0`
5. `sysctl -w net.ipv4.ip_forward=1` — now it works

## Test environment

- Hetzner CPX22 (fsn1), Ubuntu 24.04
- syfrah from `origin/main` (9afb914)

### LINKED ISSUE #1152: VPC peering veth not attached when target bridge doesn't exist yet at peering time
## Bug

When `syfrah vpc peer --from vpc-a --to vpc-b` is executed, the Raft handler creates a veth pair and tries to attach each end to the corresponding bridge. If one of the bridges does not exist yet (no VMs in that VPC on this hypervisor), the veth pair is created but the attach to the missing bridge fails silently. The veth ends up dangling — not attached to any bridge.

Later, when a VM is created in the missing VPC and the bridge appears, the reconciler sees the veth already exists and skips creation, only re-applying nftables rules. The veth is never retroactively attached to the new bridge.

## Impact

- Peering dataplane is silently broken
- Must unpeer + re-peer after VMs in both VPCs are running to get connectivity

## Root cause

`raft_handler.rs` line 467-468: the code checks `has_a || has_b` (at least one bridge exists) before calling `create_veth_peer`. It should either:
1. Only wire when BOTH bridges exist (`has_a && has_b`)
2. OR ensure the reconciler properly attaches unattached veth ends when new bridges appear

The reconciler in `reconcile.rs` line 570 checks if the veth exists and if so only re-applies nftables rules — it never checks if the veth is properly attached to its bridge.

## Reproduction

1. Create vpc-a and vpc-b
2. Create VM only in vpc-b (bridge-b exists, bridge-a does not)
3. Run `syfrah vpc peer --from vpc-a --to vpc-b`
4. Create VM in vpc-a (bridge-a now exists)
5. The peering veth exists but is not connected to bridge-a
6. Must unpeer + re-peer to fix

sacha-ops:
## Summary

Fixes three interconnected VPC networking bugs:

- **VPC isolation** (Closes #1149): Change `inet syfrah forward` chain from `policy accept` to `policy drop`. Add conntrack rules and per-bridge accept rules for intra-VPC traffic and internet egress. Cross-VPC traffic is now dropped by default unless explicit peering rules exist.

- **IP forwarding** (Closes #1150): Call `ensure_ip_forwarding()` at daemon startup alongside `br_netfilter`. Also re-enable in the reconcile loop since sysctl values don't survive reboot.

- **Peering veth guard** (Closes #1152): Change `has_a || has_b` to `has_a && has_b` in `raft_handler.rs` VpcPeer side-effect so veth pairs are only wired when both bridges exist. Added same guard to the reconcile peerings loop.

## Changes

- `layers/overlay/src/nft.rs`: Forward chain `policy drop`, conntrack rules in infra protection, new `generate_bridge_accept_rules()` function
- `layers/overlay/src/backend.rs`: New `apply_bridge_accept_rules` trait method
- `layers/overlay/src/linux.rs`: Implement `apply_bridge_accept_rules`, update all chain creation to `policy drop`
- `layers/overlay/src/mock.rs`: Implement mock `apply_bridge_accept_rules`
- `layers/overlay/src/reconcile.rs`: Re-apply bridge accept rules, check both bridges before peering, re-enable ip_forward
- `layers/overlay/src/recovery.rs`: Apply bridge accept rules after bridge recovery
- `layers/compute/src/network_setup.rs`: Apply bridge accept rules after bridge creation
- `layers/forge/src/api.rs`: Apply bridge accept rules in create_bridge_handler
- `layers/fabric/src/daemon.rs`: Enable ip_forward at startup
- `layers/fabric/src/raft_handler.rs`: `has_a && has_b` guard for peering wiring

## Test plan

- [ ] Build passes (`cargo build --workspace`)
- [ ] All tests pass (1 pre-existing unrelated failure in `binary::tests`)
- [ ] Clippy clean
- [ ] Deploy to Hetzner test servers and verify:
  - VMs in same VPC can communicate (same-bridge accept)
  - VMs in different VPCs cannot communicate without peering (policy drop)
  - After `vpc peer`: cross-VPC ping works
  - After `vpc unpeer`: cross-VPC ping fails again
  - Internet egress works from VMs

### DIFF
--- a/layers/compute/src/network_setup.rs
+++ b/layers/compute/src/network_setup.rs
@@ -280,6 +280,14 @@ impl<B: NetworkBackend + ?Sized> NetworkSetup<B> {
             .await
             .map_err(|e| ComputeError::NetworkSetup(format!("bridge creation failed: {e}")))?;
 
+        // Apply per-bridge accept rules (intra-VPC + internet egress) now
+        // that the bridge exists. With policy drop on the forward chain,
+        // traffic would be blocked without these.
+        self.backend
+            .apply_bridge_accept_rules(&bridge_name)
+            .await
+            .map_err(|e| ComputeError::NetworkSetup(format!("bridge accept rules failed: {e}")))?;
+
         self.backend
             .attach_to_bridge(&vxlan_name, &bridge_name)
             .await
--- a/layers/fabric/src/daemon.rs
+++ b/layers/fabric/src/daemon.rs
@@ -1201,6 +1201,13 @@ pub async fn run_daemon(
                                 .map_err(|e| e.to_string())
                         });
 
+                        // Enable IP forwarding so the kernel routes packets
+                        // between VPC bridges and to the internet. Without
+                        // this, cross-subnet and NAT traffic silently fails.
+                        if let Err(e) = syfrah_overlay::ensure_ip_forwarding() {
+                            warn!("compute: failed to enable ip_forward: {e}");
+                        }
+
                         // Enable bridge netfilter hooks so nftables sees
                         // bridged VM traffic (same-bridge VM-to-VM).  Must
                         // run before any bridge or nftables setup.
--- a/layers/fabric/src/raft_handler.rs
+++ b/layers/fabric/src/raft_handler.rs
@@ -460,13 +460,14 @@ impl LayerHandler for RaftOrgHandler {
                                 let peering_id = format!("{}-{}", va.id.0, vb.id.0);
                                 let backend_guard = self.network_backend.read().await;
                                 if let Some(ref backend) = *backend_guard {
-                                    // Only wire data plane on hypervisors that have at
-                                    // least one of the peered VPC bridges locally. If
-                                    // neither bridge exists, no local VMs are in either
-                                    // VPC and there is nothing to connect.
+                                    // Only wire data plane when BOTH bridges exist
+                                    // locally. If only one exists, the veth would be
+                                    // created but one end cannot attach — resulting in
+                                    // a dangling interface. The reconcile loop will
+                                    // wire the peering once the second bridge appears.
                                     let has_a = backend.link_exists(&bridge_a).await;
                                     let has_b = backend.link_exists(&bridge_b).await;
-                                    if has_a || has_b {
+                                    if has_a && has_b {
                                         if let Err(e) = syfrah_overlay::veth_peer::create_veth_peer(
                                             backend.as_ref(),
                                             &peering_id,
@@ -488,7 +489,8 @@ impl LayerHandler for RaftOrgHandler {
                                     } else {
                                         tracing::debug!(
                                             "skipping peering data plane for {from}<->{to}: \
-                                             no local bridges on this hypervisor"
+                                             need both bridges (has_a={has_a}, has_b={has_b}); \
+                                             reconcile loop will wire when ready"
                                         );
                                     }
                                 }
--- a/layers/forge/src/api.rs
+++ b/layers/forge/src/api.rs
@@ -1588,6 +1588,12 @@ async fn create_bridge_handler(
         .as_ref()
         .map(|gt| gt.register(&resource_id));
 
+    // Apply per-bridge accept rules (intra-VPC + internet egress) so
+    // that traffic is not dropped by the forward chain's policy drop.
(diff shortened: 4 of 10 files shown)


sacha-ops:
## QA — Nova

**QA: PASS**

Verified the following:

1. **Build**: `cargo build --workspace` succeeds
2. **Tests**: 518 passed, 1 pre-existing failure (unrelated `binary::tests::resolve_binary_explicit_path_not_executable_falls_through`)
3. **Clippy**: Clean, no warnings
4. **Format**: Clean

Code review verification:
- All `policy accept` references in `inet syfrah forward` chain creation are changed to `policy drop` (nft.rs line 40, linux.rs apply_vm_rules and apply_peering_rules)
- Conntrack rules added in `generate_infra_protection()` after infra port blocks
- `generate_bridge_accept_rules()` generates same-bridge accept + internet egress rules
- All bridge creation paths (network_setup, reconcile, recovery, forge API) call `apply_bridge_accept_rules`
- `ensure_ip_forwarding()` called at daemon startup and in reconcile loop
- `has_a || has_b` changed to `has_a && has_b` in raft_handler.rs
- Reconcile peerings now guards on both bridges existing

Full E2E testing requires Hetzner deployment (post-merge).

PR #1167, opened 2026-04-03
fix: prevent daemon exit after in-process Raft init

### LINKED ISSUE #1163: [Bug] Auto-join Raft daemon restart sometimes fails — new process doesn't fully start
## Bug
During auto-join Raft (#1122), after the daemon writes the raft_init sentinel and triggers a restart, the new daemon process sometimes doesn't fully start. The operator must manually run `syfrah fabric start`.

Observed on 1 out of 3 Hetzner servers during testing. The other 2 restarted cleanly.

## Root cause
The double-fork daemon restart pattern (stop + start) has a race condition. The old daemon may not fully release the control socket or redb locks before the new one tries to acquire them.

## Fix
Add a retry loop in the restart logic: after `fabric stop`, wait for the PID file to be removed and the control socket to be released before `fabric start`. Timeout after 10s.

Or: instead of stop+start, use exec() to replace the current process (no gap between old and new daemon).

sacha-ops:
## Summary

- Fixes the daemon shutting down immediately after in-process Raft initialization
- The auto-join task is the `raft_task` in the main `tokio::select!` loop — when it returned after starting Raft, the select treated it as an unexpected task exit and shut the daemon down
- Fix: park the task with `std::future::pending()` so it never completes; it gets cancelled naturally during normal daemon shutdown
- Builds on PR #1166 which introduced the in-process Raft init

## Test plan

- [ ] `cargo build && cargo clippy && cargo test` pass
- [ ] 3-node Hetzner test: S2/S3 join, Raft auto-join succeeds, daemon PID stays the same
- [ ] Repeat 3x to verify reliability

Closes #1163

### DIFF
--- a/layers/fabric/src/daemon.rs
+++ b/layers/fabric/src/daemon.rs
@@ -2748,7 +2748,14 @@ pub async fn run_daemon(
                                 "raft: joined and started in-process. No daemon restart needed. \
                                  PID unchanged."
                             );
-                            return;
+
+                            // Do NOT return — the auto-join task is the `raft_task`
+                            // in the main `tokio::select!` loop. If it completes,
+                            // the daemon interprets it as an unexpected exit and
+                            // shuts down. Instead, park this task forever (it will
+                            // be cancelled when the daemon shuts down normally).
+                            std::future::pending::<()>().await;
+                            unreachable!();
                         }
                         Ok(resp) => {
                             let status = resp.status();


sacha-ops (commented):
**Review 1: LGTM**

Root cause identified correctly: the auto-join spawned task is assigned to `raft_task`, which is awaited in the main `tokio::select!` loop. When the task completed (returned after successfully starting Raft), the select treated it as an unexpected exit and shut the daemon down.

The fix uses `std::future::pending::<()>().await` to park the task indefinitely. This is the correct pattern — the task will be cancelled when the tokio runtime shuts down during normal SIGTERM/SIGINT handling.

The `unreachable!()` after `pending().await` is a nice touch — makes it explicit that this line should never execute.

Issue #1172, opened 2026-04-03
[P2] Install script missing wireguard module + prerequisites


sacha-ops:
## Bug
The install script doesn't install wireguard-tools or load the wireguard kernel module. Without `modprobe wireguard`, the daemon can't create the syfrah0 interface.

Also missing: crun, nftables, iproute2, and the /opt/syfrah directories.

## Fix
The install script should either:
1. Install prerequisites automatically (wireguard-tools, crun, nftables, iproute2)
2. Or check for them and warn clearly: "Missing: wireguard. Install with: apt install wireguard-tools"
3. Create /opt/syfrah/{images,kernels,instances} and /run/syfrah/vms
4. `modprobe wireguard` after install

Issue #1171, opened 2026-04-03
[P1] NAT Gateway stuck at Pending with 0.0.0.0 IP — no internet from VMs


sacha-ops:
## Bug
`syfrah nat-gw create` creates the NAT GW record but it stays at Pending with public_ip 0.0.0.0. The nftables masquerade rule is never applied. VMs can ping their gateway but not the internet.

## Root cause
On the Raft path, the NAT GW creation goes through the state machine which stores the record but doesn't trigger the nftables side-effect. The side-effect (detect public IP, apply masquerade, transition to Active) only runs in the direct-write path.

## Fix
After Raft commits CreateNatGw, the local Forge must detect the public IP and apply the masquerade rule. Either:
1. Add a post-Raft side-effect in the handler (like VPC peering)
2. Or let the reconcile loop detect NAT GWs in Pending state and activate them

Issue #1170, opened 2026-04-03
[P1] Container VMs have no sshd — SSH impossible after vm create


sacha-ops:
## Bug
VMs created with `--ssh-key` accept the flag but the alpine-3.20 container image has no openssh installed. The container runs `sleep infinity` as PID 1. SSH is impossible.

This worked on previous test servers because they had a patched `runtime_container.rs` that injected sshd setup into the rootfs. That code may have been lost in a merge or the container init script isn't running.

## Fix
Verify that `inject_ssh_setup()` in `runtime_container.rs` is called during container creation. It should write authorized_keys, install openssh-server, and start sshd. If the function exists but isn't called, wire it. If it was lost, restore it.

Issue #1169, opened 2026-04-03
[P1] hypervisor register returns 'Error: unexpected response: Ok' despite success


sacha-ops:
## Bug
`syfrah hypervisor register --region eu --zone fsn1` succeeds (hypervisor IS registered) but the CLI prints "Error: unexpected response: Ok" and exits with code 1.

## Root cause
The Raft handler returns `OrgResponse::Ok` but the CLI expects a `HypervisorResponse::Hypervisor(...)` response. The response type mismatch causes the CLI to treat it as an error.

## Fix
Either change the Raft handler to return the proper response type, or make the CLI accept `Ok` as a success for register operations.

PR #1173, opened 2026-04-03
fix: 5 release bugs for v1.14.2

### LINKED ISSUE #1168: [P0] Daemon exits after 5s in bootstrap mode — breaks Quick Start
## Bug
The daemon shuts down ~5 seconds after `fabric init` when no peers exist and controlplane is not initialized. The auto-join Raft task completes (no peers to probe), and the `tokio::select!` treats this as an exit signal.

## Impact
The README Quick Start tells users to: init fabric → peering start → join peers → THEN init controlplane. But the daemon dies before peers can join.

## Reproduction
```bash
syfrah fabric init --name cloud --node-name hv-1 --endpoint 1.2.3.4:51820
# Daemon starts, exits after ~5s
```

## Fix
The auto-join Raft task must NOT exit when no peers are found. It should park itself with `std::future::pending()` (like the fix in PR #1167) or loop and retry periodically. When no Raft cluster is detected, the task should sleep and retry every 30s instead of returning.

Also: `syfrah controlplane init` should work WHILE the daemon is running (no need for stop+start).

sacha-ops:
## Summary

- **#1168 P0: Daemon exits in bootstrap mode** — Auto-join task now loops forever with 30s retries instead of returning after 5 attempts. When no peers or no Raft cluster is found, the task sleeps and retries. The daemon stays alive.
- **#1169: Hypervisor register "Error: unexpected response: Ok"** — `run_register()` now accepts `HypervisorResponse::Ok` as success and prints "Hypervisor registered."
- **#1170: Container VMs have no sshd** — Verified already wired from PR #858. `inject_ssh_setup` is called during `ContainerRuntime::create()` when SSH key is provided.
- **#1171: NAT GW stuck at Pending** — Overlay reconcile loop (30s interval) now detects Pending NAT GWs, detects public IP, applies masquerade via `backend.apply_nat()`, and transitions to Active.
- **#1172: Install script missing prerequisites** — Added wireguard-tools install, nftables install, WireGuard module load, crun check, and creates all required directories.

## Files changed

- `layers/fabric/src/daemon.rs` — Auto-join retry loop + NAT GW reconciliation
- `layers/org/src/cli/hypervisor.rs` — Accept Ok response for register
- `scripts/install.sh` — Prerequisites + directories

## Test plan

- [ ] `cargo build --workspace` passes
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [ ] `cargo test --workspace` — 518 pass, 1 pre-existing failure (unrelated)
- [ ] Quick Start on 3 fresh Hetzner servers after release

Closes #1168 #1169 #1170 #1171 #1172

### DIFF
--- a/layers/fabric/src/daemon.rs
+++ b/layers/fabric/src/daemon.rs
@@ -2103,203 +2103,220 @@ pub async fn run_daemon(
                 // Give the mesh a few seconds to settle after startup.
                 tokio::time::sleep(std::time::Duration::from_secs(5)).await;
 
-                // Skip if raft_initialized sentinel already exists (race with
-                // manual `syfrah controlplane join`).
-                if auto_join_syfrah_dir.join("raft_initialized").exists() {
-                    info!("raft: auto-join skipped — already initialized");
-                    return;
-                }
-
-                // Load current peers from fabric state.
-                let peers = match crate::store::load() {
-                    Ok(state) => state.peers,
-                    Err(_) => {
-                        info!("raft: auto-join skipped — no fabric state");
-                        return;
-                    }
-                };
-
-                if peers.is_empty() {
-                    info!("raft: no peers in mesh — staying in bootstrap mode");
-                    return;
-                }
-
-                // Derive our node ID the same way controlplane does.
-                let node_id = {
-                    use std::hash::{Hash, Hasher};
-                    let mut hasher = std::collections::hash_map::DefaultHasher::new();
-                    auto_join_node_name.hash(&mut hasher);
-                    hasher.finish()
-                };
-                let node_addr = format!("[{auto_join_ipv6}]:7200");
-
-                let probe_client = match reqwest::Client::builder()
-                    .timeout(std::time::Duration::from_secs(5))
-                    .build()
-                {
-                    Ok(c) => c,
-                    Err(e) => {
-                        warn!("raft: auto-join failed to build HTTP client: {e}");
-                        return;
-                    }
-                };
-                let join_client = match reqwest::Client::builder()
-                    .timeout(std::time::Duration::from_secs(30))
-                    .build()
-                {
-                    Ok(c) => c,
-                    Err(e) => {
-                        warn!("raft: auto-join failed to build HTTP client: {e}");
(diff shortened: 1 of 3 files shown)


sacha-ops:
## Review — Milo (reliability/correctness)

### Summary
Verified all changes against the issue descriptions:

1. **#1168 fix**: The daemon no longer exits when auto-join fails. The outer loop retries every 30s with fresh peer lists. The backoff array is still used within each cycle for probe retries. When sentinel is detected mid-loop, the task parks forever via `pending()`. Correct.

2. **#1169 fix**: The `run_register` match now handles both `Hypervisor(hv)` and `Ok` variants. This matches the Raft handler at line 155 of `raft_hypervisor_handler.rs` which returns `HypervisorResponse::Ok`.

3. **#1170**: Already implemented. `inject_ssh_setup` is called at line 1044 of `runtime_container.rs`. The OCI config uses `/syfrah-init.sh` when SSH key is present. No code change needed.

4. **#1171 fix**: The overlay reconcile loop now activates Pending NAT GWs. Uses `backend.apply_nat()` which runs nftables commands. Falls back to Failed on error. Public IP detected via `ip route get 8.8.8.8`.

5. **#1172 fix**: Install script now creates all required directories and installs wireguard-tools + nftables. The `modprobe wireguard` call ensures the kernel module is loaded.

### Verdict

**Verdict: APPROVE**

sacha-ops:
## QA — Ren

### Verification
- `cargo build --workspace` — PASS
- `cargo fmt -- --check` — PASS
- `cargo clippy --workspace --all-targets -- -D warnings` — PASS (0 warnings)
- `cargo test --workspace` — 518 pass, 1 pre-existing failure (binary::tests::resolve_binary_explicit_path_not_executable_falls_through — exists on main)

### Code review verification
- daemon.rs: auto-join loop structure is correct, no early returns that would kill daemon
- hypervisor CLI: Ok variant handled alongside Hypervisor variant
- install.sh: prerequisites installed before verification step
- NAT GW: reconcile loop properly activates Pending gateways

**QA: PASS**

Issue #1179, opened 2026-04-03
Define VolumeDesiredState and VolumeObservedState types


sacha-ops:
## Context
ADR-006 §5 — Desired vs Observed state model.

## Scope
Add to `layers/core/src/`:
- `VolumeDesiredState` enum: Available, AttachedTo { vm_id, hypervisor_id }, Deleted
- `VolumeDesiredRecord` struct: target + generation
- `VolumeObservedState` struct: nbd_connected, cache_bytes, dirty_bytes, ch_attached, last_observed
- `VolumeReportedState` enum: Creating, Available, Attaching, Attached, Detaching, Resizing, Deleting, Deleted, Error
- Function `derive_reported_state(desired: &VolumeDesiredState, observed: &VolumeObservedState) -> VolumeReportedState` implementing the mapping table from §5

## Acceptance criteria
- All types compile with Serialize/Deserialize
- `derive_reported_state()` unit tests cover all 9 reported states
- State diagram transitions match ADR-006 §5 exactly

Ref: ADR-006 §5

PR #1216, opened 2026-04-03
feat(core): define Volume and Snapshot storage types

### LINKED ISSUE #1178: Define Volume and Snapshot core types in syfrah-core
## Context
ADR-006 §4, §6 — Volume resource model and Snapshot resource model.

## Scope
Add the following types to `layers/core/src/`:
- `Volume` struct with all fields from ADR-006 §4 (id, name, size_gb, desired_state, attached_to, hypervisor_id, s3_prefix, encryption_key_id, snapshot_ids, org/project/env, placement_generation, volume_type, timestamps)
- `VolumeId`, `SnapshotId` newtypes
- `VolumeType` enum (Root, Data)
- `Snapshot` struct from §6 (id, source_volume_id, state, size_gb, s3_prefix, org/project/env, timestamps)
- `SnapshotState` enum (Creating, Available, Deleting, Deleted)
- Derive: Debug, Clone, Serialize, Deserialize, PartialEq, Eq

## Acceptance criteria
- All types compile and have Serialize/Deserialize
- Unit tests for serialization round-trip
- No I/O, no async — pure types only (syfrah-core convention)

Ref: ADR-006 §4, §6

sacha-ops:
## Summary
- Add Volume, Snapshot, VolumeId, SnapshotId, VolumeType, SnapshotState types to syfrah-core
- Include VmId, HypervisorId, OrgId, ProjectId, EnvId newtypes for struct fields
- Minimal VolumeDesiredState enum (full state model in #1179)
- Pure types with Serialize/Deserialize, no I/O
- Unit tests for serialization round-trip on all types

Closes #1178

### DIFF
--- a/layers/controlplane/src/commands.rs
+++ b/layers/controlplane/src/commands.rs
@@ -4,6 +4,67 @@ use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
 
+// ---------------------------------------------------------------------------
+// Storage types (inline until syfrah-core #1178 lands proper typed IDs)
+// ---------------------------------------------------------------------------
+
+/// Volume type: root volumes are tied to a VM lifecycle, data volumes are independent.
+// TODO: Replace with syfrah_core::VolumeType once #1178 lands.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum VolumeType {
+    /// Root volume — auto-created with a VM, deleted when the VM is deleted.
+    Root,
+    /// Data volume — created independently, lifecycle is not tied to any VM.
+    Data,
+}
+
+/// Per-region S3 storage configuration replicated through Raft.
+///
+/// # Security notes
+/// - `s3_access_key` and `s3_secret_key` ARE stored in Raft so that all nodes
+///   in the region know how to reach the S3 bucket.
+/// - The `encryption_passphrase` is NOT stored in Raft. It is kept locally on
+///   each hypervisor (file with 0600 permissions) and never replicated via
+///   consensus. See ADR-006 §9.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct StorageConfig {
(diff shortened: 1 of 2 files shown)


sacha-ops:
## Review — Kira

**Verdict: APPROVE**

Clean, minimal types PR. No I/O, no unsafe, no panicking code paths. Serde roundtrip tests cover all variants. A few security observations, all non-blocking for a types-only crate:

1. **`encryption_key_id` in `Debug` output** (`storage.rs:127`): The `Volume` struct derives `Debug`, which will print `encryption_key_id` in logs. The field appears to be a key *reference* (not material), and the test fixture uses `"key-default"`, so this is acceptable. If this ever holds an ARN, KMS path, or anything more sensitive, a custom `Debug` impl that redacts it would be warranted -- same pattern as `MeshSecret::Debug` in `secret.rs:191`.

2. **`s3_prefix` in `Debug` output** (`storage.rs:125`, `storage.rs:170`): Both `Volume` and `Snapshot` expose S3 paths via derived `Debug`. Low risk today, but worth remembering that structured logging will broadcast internal bucket paths. Not blocking.

3. **No format validation on newtype IDs** (`storage.rs:11`, `storage.rs:21`, etc.): All ID newtypes (`VolumeId`, `SnapshotId`, `VmId`, etc.) have `pub` inner fields and accept any `String` via both construction and deserialization. The doc comments describe formats like `vol-{ulid}` but nothing enforces them. For a core types crate this is fine -- validation belongs at the API boundary -- but ensure that the gateway/admin layers validate ID formats before constructing these types. If a malicious or malformed ID reaches storage logic, path traversal via `s3_prefix` construction from an unvalidated `VolumeId` becomes the real threat. Track this as a follow-up when the storage service layer lands.

4. **Positive notes**: No `Display` on `Volume`/`Snapshot` (only on IDs) avoids accidentally dumping full structs into user-facing output. Encryption key ID is separate from key material. Timestamps are `u64` (no timezone/parsing issues).

Good to merge.

PR #1224, opened 2026-04-03
feat(storage): ZeroFS binary management and version pinning

### LINKED ISSUE #1186: ZeroFS binary management: bundle, install, version pinning
## Context
ADR-006 §14 — ZeroFS binary management, same pattern as Cloud Hypervisor.

## Scope
- Add ZeroFS binary to release tarball alongside syfrah and cloud-hypervisor
- Install to `/usr/local/lib/syfrah/zerofs`
- Update `scripts/install.sh` to include zerofs
- Binary resolution order: PATH → /usr/local/lib/syfrah/zerofs → bundled
- Version pinning: store expected ZeroFS version, warn on mismatch
- `syfrah storage version` command to display ZeroFS version

## Acceptance criteria
- ZeroFS binary included in release tarball
- install.sh installs zerofs to correct location
- Binary resolution works (same pattern as Cloud Hypervisor in compute layer)
- Version mismatch produces warning, not hard failure

Ref: ADR-006 §14

sacha-ops:
## Summary
- ZeroFS binary resolution following Cloud Hypervisor pattern (PATH → `/usr/local/lib/syfrah/zerofs` → bundled)
- Compile-time version pinning via `ZEROFS_VERSION` file at repo root
- Version checking with mismatch warnings
- `syfrah storage version` CLI command to display ZeroFS version and path
- Updated `install.sh` to install zerofs binary
- Updated release workflow to download and bundle zerofs in tarballs

Closes #1186

### DIFF
--- a/layers/storage/src/binary.rs
+++ b/layers/storage/src/binary.rs
@@ -30,8 +30,8 @@ pub enum ZerofsError {
 ///
 /// Resolution order:
 /// 1. `explicit` path (if provided, must exist and be executable)
-/// 2. `zerofs` on `$PATH` (via `which`)
-/// 3. `/usr/local/lib/syfrah/zerofs` (standard install location)
+/// 2. `/usr/local/lib/syfrah/zerofs` (standard install location)
+/// 3. `zerofs` on `$PATH` (via `which`)
 ///
 /// Returns the first match or an error if none found.
 pub fn resolve_binary(explicit: Option<&Path>) -> Result<PathBuf, ZerofsError> {
@@ -48,7 +48,13 @@ pub fn resolve_binary(explicit: Option<&Path>) -> Result<PathBuf, ZerofsError> {
         // File exists but is not executable — fall through.
     }
 
-    // 2. Search $PATH via `which`
+    // 2. Standard installation path
+    let installed = PathBuf::from("/usr/local/lib/syfrah/zerofs");
+    if is_executable(&installed) {
+        return Ok(installed);
+    }
+
+    // 3. Search $PATH via `which`
     if let Ok(output) = std::process::Command::new("which").arg("zerofs").output() {
         if output.status.success() {
             let path_str = String::from_utf8_lossy(&output.stdout);
@@ -57,12 +63,6 @@ pub fn resolve_binary(explicit: Option<&Path>) -> Result<PathBuf, ZerofsError> {
                 return Ok(path);
             }
         }
-    }
-
-    // 3. Standard installation path
-    let installed = PathBuf::from("/usr/local/lib/syfrah/zerofs");
-    if is_executable(&installed) {
-        return Ok(installed);
     }
 
     Err(ZerofsError::NotFound {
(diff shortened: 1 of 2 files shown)


sacha-ops:
## Review — Ren

**Verdict: REQUEST_CHANGES**

Solid implementation overall — clean module structure, good docstrings, and the test coverage for the new binary.rs is a good start. A few things need fixing before merge:

---

### 1. Resolution order diverges from Cloud Hypervisor pattern (bug)

The module docstring says it "follows the same pattern as Cloud Hypervisor (`layers/compute/src/binary.rs`)" but the resolution order is actually different:

| Priority | Cloud Hypervisor | ZeroFS (this PR) |
|----------|-----------------|------------------|
| 1 | explicit path | explicit path |
| 2 | `/usr/local/lib/syfrah/cloud-hypervisor` | `$PATH` (via `which`) |
| 3 | `$PATH` (via `which`) | `/usr/local/lib/syfrah/zerofs` |

The installed binary at `/usr/local/lib/syfrah/zerofs` should take priority over whatever happens to be on `$PATH`. The current order means a stale or unrelated `zerofs` in `/usr/bin` would shadow the version-pinned installed binary. Swap steps 2 and 3 to match CH.

**File:** `layers/storage/src/binary.rs`, lines 51-66

---

### 2. Missing test coverage compared to CH reference

The CH binary module has notably more test coverage. Missing from this PR:

- **`verify_version` tests** — no tests for `verify_version()` at all (match case, mismatch case). CH has both.
- **`resolve_binary(None)` when nothing is found** — CH tests the "none found" path explicitly.
- **`parse_version_output` with patch/pre-release versions** — e.g., `"zerofs v0.1.2-rc1"` — CH tests `check_version_with_patch_version`.
- **`check_version` with empty output** — CH tests this case.
- **Priority/precedence test** — CH has `resolve_binary_explicit_takes_priority`. Worth adding one that confirms explicit > installed > PATH ordering.

These are the reliability-critical paths. Binary resolution bugs are silent and painful to debug in production.

---

### 3. `run_storage` reimplements version mismatch logic instead of calling `verify_version`

In `cli/mod.rs`, the `StorageCommand::Version` handler manually compares `ver != pinned` and prints a warning. The `verify_version()` function in `binary.rs` already does exactly this. Use it:

```rust
if let Err(warning) = crate::binary::verify_version(&path) {
    eprintln!("warning: {warning}");
}
```

This avoids duplicating the comparison logic and keeps the CLI consistent with any future callers.

**File:** `layers/storage/src/cli/mod.rs`, lines ~183-195

---

### 4. Minor: release workflow silently skips zerofs on download failure

In `.github/workflows/release.yml`, a failed ZeroFS download emits `::warning::` but the step succeeds. This means a release tarball can ship without zerofs and no one notices unless they check workflow logs. Consider whether this should be a hard failure for Linux targets, or at minimum ensure the warning is surfaced more visibly (e.g., in the release notes or as a job annotation).

---

### Summary

The resolution order bug (#1) is the blocking issue — it's a correctness problem that contradicts the documented behavior. The missing tests (#2) and duplicated logic (#3) are important for reliability but lower severity. #4 is a nice-to-have.

PR #1227, opened 2026-04-03
feat(storage): implement VolumeMgr for ZeroFS process lifecycle

### LINKED ISSUE #1187: Implement VolumeMgr in Forge: start/stop ZeroFS processes

sacha-ops:
## Summary
- Add `VolumeMgr` struct to `layers/storage/src/volume_mgr.rs` that manages ZeroFS child processes per volume
- `start_volume`: resolves ZeroFS binary (via `binary.rs`), spawns process with S3/cache/encryption args, waits for NBD device, tracks child
- `stop_volume`: SIGTERM with 10s grace period, then SIGKILL if needed
- `is_running` / `list_active` for querying state
- `reap_exited` for process supervision (detect crashes)
- 6 unit tests covering duplicate start prevention, stop unknown volume, process termination, crash detection, and running process preservation

## Test plan
- [x] `cargo test -p syfrah-storage` — all 64 tests pass (6 new)
- [ ] Integration test with real ZeroFS binary (deferred to follow-up)

Closes #1187

### DIFF
--- a/layers/storage/src/lib.rs
+++ b/layers/storage/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod api;
 pub mod binary;
 pub mod cli;
+pub mod nbd;
 pub mod volume_mgr;
 
 pub use api::{
--- a/layers/storage/src/nbd.rs
+++ b/layers/storage/src/nbd.rs
@@ -1,0 +1,568 @@
+//! NBD (Network Block Device) allocator for `/dev/nbdN` device management.
+//!
+//! This module tracks which `/dev/nbdN` slots are in use so that ZeroFS
+//! instances each get a unique block device. The allocator does **not** manage
+//! the kernel NBD connection itself — ZeroFS takes a device path and handles
+//! the `nbd-client`/ioctl lifecycle.
+//!
+//! # Edge cases handled
+//!
+//! - Kernel module not loaded (`modprobe nbd` required)
+//! - All device slots exhausted
+//! - Device file missing from `/dev` even though the slot number is in range
+//! - Concurrent allocation via interior `Mutex`
+
+use std::collections::HashSet;
+use std::fmt;
+use std::path::{Path, PathBuf};
+use std::sync::Mutex;
+
+/// Default maximum number of NBD devices if the kernel module parameter cannot
+/// be read.
+const DEFAULT_MAX_DEVICES: usize = 16;
+
+/// Sysfs path where the `nbd` module's `nbds_max` parameter is exposed.
+const NBDS_MAX_PATH: &str = "/sys/module/nbd/parameters/nbds_max";
+
+// ---------------------------------------------------------------------------
+// Error type
+// ---------------------------------------------------------------------------
+
+/// Errors returned by NBD allocator operations.
+#[derive(Debug, thiserror::Error)]
+pub enum NbdError {
+    /// The `nbd` kernel module is not loaded and `modprobe` failed.
+    #[error("failed to load nbd kernel module: {reason}")]
+    ModuleLoadFailed { reason: String },
+
+    /// All `/dev/nbdN` device slots are currently in use.
+    #[error("all {max} NBD device slots are in use")]
+    AllDevicesBusy { max: usize },
+
+    /// The `/dev/nbdN` device file does not exist on disk.
+    #[error("device file does not exist: {path}")]
+    DeviceFileMissing { path: String },
+
+    /// Attempted to release a device that was not allocated.
+    #[error("device /dev/nbd{n} is not currently allocated")]
+    NotAllocated { n: usize },
+
+    /// An I/O error occurred while probing the system.
+    #[error("I/O error: {0}")]
+    Io(#[from] std::io::Error),
+}
+
+// ---------------------------------------------------------------------------
+// NbdDevice
+// ---------------------------------------------------------------------------
+
+/// A handle representing an allocated `/dev/nbdN` device.
+///
+/// This is a plain data struct — dropping it does **not** automatically release
+/// the slot. Call [`NbdAllocator::release`] explicitly when the device is no
+/// longer needed.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct NbdDevice {
+    /// The device number (the `N` in `/dev/nbdN`).
+    pub number: usize,
+}
+
+impl NbdDevice {
+    /// The full device path, e.g. `/dev/nbd0`.
+    ///
+    /// Derived from `number` so there is only one source of truth.
+    pub fn path(&self) -> PathBuf {
+        PathBuf::from(format!("/dev/nbd{}", self.number))
+    }
+}
+
+impl fmt::Display for NbdDevice {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "/dev/nbd{}", self.number)
+    }
+}
+
+// ---------------------------------------------------------------------------
+// System interaction trait (for testing)
+// ---------------------------------------------------------------------------
+
+/// Trait abstracting system-level operations so unit tests can run without real
+/// kernel devices.
+pub trait NbdSystem: Send + Sync {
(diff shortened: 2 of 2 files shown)


sacha-ops:
## Review — Nova

Reviewed for: process lifecycle efficiency, resource cleanup, lock contention, scalability.

### Findings

**1. NBD index allocation never reclaims freed slots (medium)**
`next_nbd_index` is a monotonically increasing counter. When a volume is stopped and its NBD device freed, that index is never returned to the pool. On a long-running node that cycles volumes (e.g. live migration, scale-down/up), this will exhaust `/dev/nbdN` devices unnecessarily. A free-list or reuse strategy is needed.

**2. Secret key passed as CLI argument — visible in `/proc/{pid}/cmdline` (high)**
`--s3-secret-key` and `--encryption-key` are passed as positional CLI args to the child process. On Linux any user can read `/proc/<pid>/cmdline`. These secrets should be passed via environment variables or a temporary file descriptor (e.g. `--secret-key-fd 3`) instead.

**3. No concurrency safety — `&mut self` everywhere (medium)**
Every public method takes `&mut self`. In a real runtime, multiple tasks will want to start/stop/query volumes concurrently. This design forces external callers to wrap the entire `VolumeMgr` in a single `Mutex`, which serializes all operations including the blocking `wait_for_nbd` (up to 30s). Consider:
- Using a `DashMap` or `Arc<Mutex<...>>` per-entry so start/stop on different volumes can proceed in parallel.
- At minimum, splitting the map lookup from the long-running NBD wait so the lock isn't held for 30 seconds.

**4. `wait_for_nbd` is called while holding mutable borrow but before map insertion (acceptable)**
Currently the NBD wait happens *before* the process is inserted into the map, which means a crash during the wait window leaves an orphaned child process with no tracking. If the spawn succeeds but the NBD device never appears, the child process is leaked — nobody calls `kill()` on it. Add cleanup on the timeout path:
```rust
Err(_) => {
    let _ = child.kill().await;
    Err(VolumeMgrError::NbdTimeout(NBD_WAIT_TIMEOUT))
}
```

**5. `kill_on_drop(false)` in `start_volume` — intentional orphan risk (low)**
The spawned child uses `kill_on_drop(false)`. If the `VolumeMgr` is dropped without calling `stop_volume` on all entries, every child process becomes an orphan. Consider implementing `Drop` for `VolumeMgr` that sends SIGTERM to all tracked children, or at least document this contract explicitly.

**6. `reap_exited` is O(n) scan over all processes (acceptable for now)**
Fine at current scale but worth noting: if we ever manage hundreds of volumes per node, a `waitpid`-based reaper (via `tokio::signal::unix`) would be more efficient than polling.

**7. `list_active` clones all keys (low)**
Returns `Vec<String>` by cloning every key. For a status/health endpoint called frequently, consider returning an iterator or `Vec<&str>`.

### Positive notes
- SIGTERM-then-SIGKILL with grace period is the correct pattern.
- Clean error types with `thiserror`.
- Tests are well-structured and cover the important edge cases (duplicate start, unknown stop, crash detection).

### Required changes before merge
Items 2 and 4 are blockers:
- **#2**: Do not pass secrets via CLI args. Use env vars at minimum.
- **#4**: Kill the child process on NBD timeout to prevent process leaks.

**Verdict: REQUEST_CHANGES**

sacha-ops:
## Review — Axel

Reviewed from a process lifecycle, production safety, and failure recovery perspective.

### Issues

**1. Encryption passphrase passed as CLI argument (CRITICAL)**
The encryption passphrase is passed via `--encryption-key` as a command-line argument. On any Linux system, `/proc/<pid>/cmdline` is world-readable, which means any user on the node can read the passphrase. This is a real production data-at-rest encryption leak. Pass it via stdin, an environment variable, or a file descriptor instead.

**2. S3 secret key also on command line (CRITICAL)**
Same problem for `--s3-secret-key`. Anyone with shell access to the node can harvest S3 credentials from `/proc`. Use environment variables (`Command::env()`) or pass via stdin.

**3. `kill_on_drop(false)` in `start_volume` — orphan risk (HIGH)**
If the `VolumeMgr` is dropped without calling `stop_volume` for every tracked volume (panic, early return, process crash), all ZeroFS children become orphans that keep running indefinitely. Since these hold NBD devices, this means resources are leaked and cannot be reclaimed without manual intervention. Either:
- Set `kill_on_drop(true)`, or
- Implement `Drop` for `VolumeMgr` that sends SIGTERM to all tracked children, or
- Document clearly why orphaning is intentional (I can imagine wanting the volume to survive a supervisor restart, but then you need a PID-file reconnect mechanism).

**4. No `Drop` implementation for `VolumeMgr` (HIGH)**
Related to above. If `VolumeMgr` is dropped (e.g., during a panic unwind), there is zero cleanup. At minimum, a `Drop` impl should log a warning if `self.processes` is non-empty, and ideally send SIGTERM to all children.

**5. NBD index allocation is not collision-safe (MEDIUM)**
`next_nbd_index` is a simple monotonic counter starting at 0. If volumes are started and stopped, indices are never reclaimed. If a second `VolumeMgr` instance is created (e.g., after a restart), it starts at 0 again and will collide with any surviving orphan processes still holding `/dev/nbd0`. Consider querying the kernel for available NBD devices or maintaining a persistent allocation map.

**6. SIGTERM race: process removed before kill confirmed (MEDIUM)**
In `stop_volume`, the process entry is removed from the map (`self.processes.remove`) *before* the SIGTERM/wait/SIGKILL sequence. If `stop_volume` fails partway (e.g., timeout + SIGKILL fails), the process is already removed from tracking, so it becomes an orphan that `reap_exited` will never find. Move the `remove` to after successful termination.

**7. No stdout/stderr capture or logging (MEDIUM)**
The spawned ZeroFS process inherits the parent's stdio. In production, you need to either:
- Redirect stdout/stderr to a log file per volume, or
- Pipe them and forward to the structured logging system.
Without this, debugging a crashed ZeroFS process in production is guesswork.

**8. `reap_exited` does not report exit codes (LOW)**
`try_wait()` returns the `ExitStatus` but it is silently discarded. Logging the exit code (and especially whether it was killed by a signal) is essential for crash diagnostics. The returned `Vec<String>` should include exit status information.

**9. No `#[cfg(unix)]` guard on SIGTERM in `stop_volume` (LOW)**
The `#[cfg(unix)]` block around the SIGTERM call means on non-Unix platforms, `stop_volume` will skip SIGTERM entirely and just wait for the grace period to expire before calling `kill()`. This is a 10-second unnecessary delay on non-Unix. Not critical since this is Linux-only in practice, but worth a comment or a compile-time assertion.

### Positives

- SIGTERM-then-SIGKILL with a grace period is the correct shutdown pattern.
- `reap_exited` for supervision/crash detection is a good design.
- Duplicate start prevention avoids double-attach to the same volume.
- Clean error types with `thiserror`.
- Tests cover the important edge cases (duplicate, unknown stop, crash detection, running preservation).

### Summary

The core lifecycle design is sound, but there are two credential-leak issues that are blockers for any production deployment, plus the orphan-on-drop risk needs addressing before this manages real infrastructure.

**Verdict: REQUEST_CHANGES**

Blocking: #1 and #2 (credential exposure via `/proc/cmdline`), #3 or #4 (orphan risk on drop), #6 (remove-before-kill race).

PR #1228, opened 2026-04-03
feat(storage): NBD device lifecycle allocator

### LINKED ISSUE #1188: NBD device lifecycle: create, connect, disconnect, destroy
## Context
ADR-006 §14 — NBD device lifecycle per volume.

## Scope
- NBD device allocation: track which /dev/nbdN devices are in use on this hypervisor
- Load nbd kernel module if not loaded (`modprobe nbd`)
- Allocate next free /dev/nbdN for new volume
- Connect: ZeroFS exposes NBD server, device appears
- Disconnect: detach from kernel, free the /dev/nbdN slot
- Destroy: ensure no lingering references after volume detach
- Handle max NBD devices limit (default 16, configurable via kernel module param)

## Acceptance criteria
- NBD devices allocated without collisions
- Device appears in /dev/ after connect
- Device removed after disconnect
- Handles kernel module loading
- Integration test: create NBD → verify /dev/nbdN exists → disconnect → verify gone

Ref: ADR-006 §14

sacha-ops:
## Summary

- Add `NbdAllocator` struct in `layers/storage/src/nbd.rs` that tracks which `/dev/nbdN` device slots are in use
- `load_module()` ensures the `nbd` kernel module is loaded via `modprobe`
- `allocate()` finds the next free `/dev/nbdN`, marks it in-use, returns an `NbdDevice` handle
- `release(device)` frees a previously allocated slot
- `is_available(n)` checks if a device exists and is not allocated
- `max_devices()` reads `nbds_max` from sysfs or defaults to 16
- Edge cases covered: module not loaded, all devices busy, device file missing, double release, concurrent access
- Testable via `NbdSystem` trait abstraction — 15 unit tests pass without real kernel devices

## Test plan

- [x] `cargo fmt` clean
- [x] `cargo test -p syfrah-storage` — all 73 tests pass (15 new NBD tests)
- [ ] CI green

Closes #1188

### DIFF
--- a/layers/storage/src/nbd.rs
+++ b/layers/storage/src/nbd.rs
@@ -65,8 +65,15 @@ pub enum NbdError {
 pub struct NbdDevice {
     /// The device number (the `N` in `/dev/nbdN`).
     pub number: usize,
+}
+
+impl NbdDevice {
     /// The full device path, e.g. `/dev/nbd0`.
-    pub path: PathBuf,
+    ///
+    /// Derived from `number` so there is only one source of truth.
+    pub fn path(&self) -> PathBuf {
+        PathBuf::from(format!("/dev/nbd{}", self.number))
+    }
 }
 
 impl fmt::Display for NbdDevice {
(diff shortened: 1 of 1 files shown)


sacha-ops:
## Review — Soren (re-review)

Verified the fix for my main concern (redundant `path` field on `NbdDevice`):

- `path: PathBuf` field is removed. `NbdDevice` now stores only `number: usize`.
- `pub fn path(&self) -> PathBuf` computes the path from `number`, eliminating the dual-source-of-truth problem.
- `Display` impl and `path()` both derive from the same `self.number` — no divergence possible.
- Tests confirm the computed path works correctly (`dev.path() == PathBuf::from("/dev/nbd0")`).

Clean fix, exactly what was needed.

**Verdict: APPROVE**

PR #1229, opened 2026-04-03
Cache configuration: SSD path, memory limit, disk limit for ZeroFS

### LINKED ISSUE #1189: Cache configuration: SSD path, memory limit, disk limit for ZeroFS
## Context
ADR-006 §10 — Two-tier cache (memory + SSD).

## Scope
- Pass cache configuration to ZeroFS on startup:
  - SSD cache path (e.g., /mnt/cache or /dev/nvme1n1)
  - SSD cache size limit in GB
  - Memory cache size limit in GB
- Per-volume cache directory: `/mnt/cache/vol-{id}/` (subdirectory per volume)
- Read cache config from StorageConfig (Raft) + per-hypervisor overrides
- Validate cache disk exists and has sufficient space before starting ZeroFS
- Report cache utilization via VolumeMgr metrics

## Acceptance criteria
- ZeroFS receives correct cache args on startup
- Per-volume cache directories created/cleaned up
- Validation: error if cache disk missing or insufficient space
- Cache size respects configured limits

Ref: ADR-006 §10

sacha-ops:
## Summary

- Add `CacheConfig` struct (ssd_path, ssd_size_gb, memory_size_gb) and `CacheDiskInfo` for disk space reporting
- Add `VolumeCacheDir` with `create_volume_cache` / `cleanup_volume_cache` for per-volume cache directories at `{ssd_path}/vol-{id}/`
- Add `validate_cache_disk` to verify cache path exists, is a directory, and report available space via `statvfs`
- Add `zerofs_cache_args` to build ZeroFS CLI arguments (`--cache-dir`, `--cache-size`, `--memory-size`)
- **Security**: volume ID validation rejects path traversal (`../`, `/`, `\`, null bytes, special characters); post-creation canonicalize check ensures resolved path stays inside SSD root

Implements ADR-006 §10 cache architecture requirements.

## Test plan

- [x] 22 unit tests covering:
  - Volume ID validation: empty, `.`, `..`, forward/back slashes, embedded `..`, null bytes, spaces, shell metacharacters
  - CacheConfig and CacheDiskInfo serde roundtrips
  - validate_cache_disk: nonexistent path, file (not dir), success on real tmpdir
  - create/cleanup volume cache: happy path, traversal rejection, idempotent cleanup
  - zerofs_cache_args: correct arg generation, traversal rejection
  - disk_space: real filesystem query, nonexistent path error
- [x] `cargo test -p syfrah-storage` — all 79 tests pass

Closes #1189

### DIFF
--- a/layers/storage/src/lib.rs
+++ b/layers/storage/src/lib.rs
@@ -14,4 +14,4 @@ pub use cache::{
     CacheDiskInfo, CacheError, VolumeCacheDir,
 };
 pub use cli::{StorageCommand, VolumeCommand};
-pub use volume_mgr::{CacheConfig, S3Config, VolumeMgr, VolumeMgrError};
+pub use volume_mgr::{S3Config, VolumeMgr, VolumeMgrError};


sacha-ops:
## Review — Ren

**Verdict: APPROVE**

Good module. The path-traversal defense is layered (validate + canonicalize post-creation), serde roundtrips are tested, cleanup is idempotent, and the `unsafe` block is correctly scoped with a clear safety comment. 22 tests covering the right edge cases.

### Issues (non-blocking)

1. **Double `vol-` prefix in directory name** — `format!("vol-{volume_id}")` with a volume ID like `vol-01JA...` produces `vol-vol-01JA...`. The tests assert this, so it appears intentional, but callers will need to know whether to pass the raw ULID or the full `vol-{ULID}` ID. Consider documenting the expected input format on `create_volume_cache` / `zerofs_cache_args`, or stripping the prefix if the caller is expected to pass the full volume ID.

2. **`used_gb` vs `available_gb` inconsistency** — `used_gb` is computed from `f_blocks - f_bfree` (includes reserved blocks) while `available_gb` uses `f_bavail` (excludes reserved blocks). This means `total_gb != used_gb + available_gb` on most filesystems. Not a bug per se, but consumers doing arithmetic on these fields will get surprising results. Consider using `f_bfree` for both, or documenting the semantics.

3. **`to_string_lossy` in `disk_space`** — `path.to_string_lossy().as_bytes()` replaces invalid UTF-8 with U+FFFD, which would produce a multi-byte CString that won't match the real path. On Unix, prefer `std::os::unix::ffi::OsStrExt::as_bytes()` on `path.as_os_str()` to preserve the raw bytes.

4. **`cleanup_volume_cache` skips canonicalize check** — `create_volume_cache` does a post-creation canonicalize containment check, but `cleanup_volume_cache` does not. The `validate_volume_id` allowlist (alphanumeric + hyphens) makes this safe in practice, but for defense-in-depth symmetry, consider adding the same check before `remove_dir_all`.

5. **Integer truncation in space check** — `available_bytes / GB` truncates, so 1.9 GB free reports as 1 GB. If `ssd_size_gb` is 2, the check fails even though there might be almost enough space. Acceptable for configuration-level granularity but worth a doc comment noting that values are floor-truncated.

### Positive observations

- Validate-then-canonicalize is the right two-layer approach for path safety
- `validate_volume_id` allowlist (alphanumeric + hyphens only) is strict enough to make traversal structurally impossible, and the canonicalize is a good belt-and-suspenders addition
- Idempotent cleanup is correct
- Tests cover all the important edge cases: empty, `.`, `..`, slashes, embedded `..`, null bytes, shell metacharacters
- `saturating_sub` for used bytes prevents underflow

All issues are non-blocking. Approve as-is; consider addressing items 1-3 in a follow-up.

sacha-ops:
## Review — Milo (re-review)

**Verdict: APPROVE**

The MUST FIX issue (double `vol-` prefix) is resolved. All three call sites — `create_volume_cache`, `cleanup_volume_cache`, and `zerofs_cache_args` — now use `config.ssd_path.join(volume_id)` directly, producing clean paths like `/mnt/cache/vol-01JA...` instead of the previous `/mnt/cache/vol-vol-01JA...`. Test assertions match. Good fix.

The SHOULD FIX (CLI unit ambiguity) and minor observations from the first review remain valid as future improvements but are non-blocking.

PR #1230, opened 2026-04-03
feat(forge): storage reconciler for volume create e2e

### LINKED ISSUE #1190: Volume create end-to-end flow: Raft → Forge reconciler → ZeroFS init
## Context
ADR-006 §4, §18 — Volume creation flow.

## Scope
Wire the full create flow:
1. CLI sends CreateVolume → daemon control socket → Raft commit
2. Raft state machine creates volume record (desired: Available)
3. Volume appears in Forge's materialized view on assigned hypervisor
4. Forge storage reconciler detects new volume
5. VolumeMgr initializes ZeroFS with S3 prefix `volumes/{id}/gen-1/`
6. ZeroFS creates S3 bucket structure (manifest, wal/, sst/)
7. Observed state updated: available, NBD disconnected (not attached yet)

## Acceptance criteria
- Volume creation from CLI propagates through Raft to Forge
- S3 prefix structure initialized correctly
- Observed state converges to match desired state
- Integration test with real ZeroFS + mock S3 (or MinIO)

Ref: ADR-006 §4, §18

sacha-ops:
## Summary

- Add `StorageReconciler` to `syfrah-forge` that bridges Raft desired state to local ZeroFS process management via `VolumeMgr`
- On each 5s tick: read volumes assigned to this hypervisor from Raft, diff against running ZeroFS processes, start missing volumes, stop orphaned/fenced ones, reap crashed processes
- Introduce `VolumeStateReader` trait for clean testability (mock Raft state in tests)
- Add `Volume = 8` variant to `ResourceType` enum for reconciliation ordering
- 11 new tests covering: no-config skip, empty state, pass counting, error reporting on missing binary, loop shutdown

## Test plan

- [x] `cargo fmt` clean
- [x] All 151 forge tests pass (including 11 new storage reconciler tests)
- [ ] Manual: verify reconciler starts ZeroFS when a volume is created via `syfrah volume create` (requires full daemon + Raft cluster)

Closes #1190

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/layers/forge/Cargo.toml
+++ b/layers/forge/Cargo.toml
@@ -30,6 +30,7 @@ anyhow.workspace = true
 uuid = { version = "1", features = ["v4"] }
 
 [dev-dependencies]
+syfrah-storage = { path = "../storage", features = ["test-helpers"] }
 tokio = { workspace = true, features = ["test-util"] }
 http-body-util = "0.1"
 tower = { version = "0.5", features = ["util"] }
--- a/layers/forge/src/storage_reconciler.rs
+++ b/layers/forge/src/storage_reconciler.rs
@@ -15,7 +15,7 @@
 //! The reconciler does NOT modify Raft state. It only reads desired state
 //! and converges local ZeroFS processes to match.
 
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::Arc;
 
@@ -181,10 +181,10 @@ impl StorageReconciler {
             }
         };
 
-        // 3. Build desired set.
+        // 3. Build desired set and running map (with generations).
         let desired_map: HashMap<String, &DesiredVolume> =
             desired.iter().map(|v| (v.id.clone(), v)).collect();
-        let running: HashSet<String> = volume_mgr.list_active().into_iter().collect();
+        let running: HashMap<String, u64> = volume_mgr.list_active().into_iter().collect();
 
         let s3 = S3Config {
             endpoint: config.s3_endpoint.clone(),
@@ -198,9 +198,39 @@ impl StorageReconciler {
             memory_size_bytes: config.cache_memory_size_bytes,
         };
 
-        // 4. Start volumes that are desired but not running.
+        // 4. Generation fencing: stop volumes running with a stale generation.
+        //    After stopping, the volume will be picked up as "desired but not
+        //    running" in step 5 and restarted with the correct generation.
         for (id, vol) in &desired_map {
-            if !running.contains(id) {
+            if let Some(&running_gen) = running.get(id.as_str()) {
+                if running_gen != vol.placement_generation {
+                    warn!(
+                        volume_id = %id,
+                        running_generation = running_gen,
+                        desired_generation = vol.placement_generation,
+                        "storage reconciler: generation mismatch, stopping stale volume"
+                    );
+                    match volume_mgr.stop_volume(id).await {
+                        Ok(()) => {
+                            report.stopped += 1;
+                        }
+                        Err(e) => {
+                            error!(
+                                volume_id = %id,
+                                error = %e,
+                                "storage reconciler: failed to stop stale-generation volume"
+                            );
+                            report.errors.push(format!("stop-stale {id}: {e}"));
+                        }
+                    }
+                }
+            }
+        }
+
+        // 5. Start volumes that are desired but not running (including those
+        //    just stopped due to generation mismatch above).
+        for (id, vol) in &desired_map {
+            if !volume_mgr.is_running(id) {
                 info!(
                     volume_id = %id,
                     generation = vol.placement_generation,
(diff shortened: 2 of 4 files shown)


sacha-ops:
## Review — Kira

**Focus:** fencing correctness, race conditions, authorization

### Findings

**1. [CRITICAL] Generation-mismatch fencing is documented but not implemented**

The module doc (lines 13-14) states:
> Generation mismatch -> stop_volume (stale placement, fencing)

But `reconcile_once` never checks generations of already-running volumes. The code only checks membership in the desired set vs. running set — it never compares `placement_generation` of the desired volume against the generation stored in the running `VolumeProcess`.

This means: if a volume is live-migrated (new generation issued in Raft), the old hypervisor will keep serving the stale generation indefinitely as long as the volume remains in its desired set. This is a split-brain / data-corruption vector.

**Fix:** After the start loop, iterate running volumes that ARE in the desired map but whose `placement_generation` differs from the tracked generation in `VolumeMgr`. Stop and restart them. This requires `VolumeMgr` to expose the generation of running volumes (e.g., `pub fn running_generation(&self, volume_id: &str) -> Option<u64>`).

**2. [MEDIUM] `list_active` returns only IDs — no generation metadata**

`VolumeMgr::list_active()` returns `Vec<String>`. The reconciler has no way to detect generation drift even if it wanted to. The API needs to return `(volume_id, generation)` pairs for the reconciler to do generation-aware diffing.

**3. [LOW] `encryption_passphrase` stored as plaintext `String` in struct**

`StorageReconciler` holds `encryption_passphrase: String` for the entire process lifetime. Consider using `secrecy::SecretString` (or zeroize-on-drop) so the passphrase does not linger in memory or appear in debug output. Not blocking, but worth a follow-up.

**4. [LOW] `last_report` uses `std::sync::Mutex` in async context**

`self.last_report.lock().unwrap()` is called inside `reconcile_once` which is `async`. A `std::sync::Mutex` held across `.await` would block the executor, but here the lock is released before any `.await`, so this is safe in practice. However, if the code evolves, it could become a footgun. Consider `tokio::sync::Mutex` or keep the pattern as-is with a comment explaining the invariant.

**5. [INFO] Tests are mostly no-op due to private `VolumeMgr` internals**

`reconcile_stops_orphaned_volume` spawns a `sleep` process but cannot inject it into `VolumeMgr` (private `processes` field), so it asserts `stopped == 0` — not actually testing the stop path. `reconcile_reaps_dead_processes` similarly asserts 0 reaps. These tests document intent but don't validate behavior. Consider adding a `VolumeMgr::inject_test_process` (cfg(test)) or integration tests.

### Summary

The reconciler structure is solid — clean trait abstraction, proper shutdown handling, and the start/stop diff logic is correct for the membership check. However, the **generation fencing path is missing**, which is the security-critical piece for preventing split-brain on volume migration. The doc comment promises it but the code doesn't deliver.

**Verdict: REQUEST_CHANGES**

Must fix before merge:
- Implement generation-mismatch detection and restart (finding 1)
- Expose running generation from `VolumeMgr` (finding 2)

sacha-ops:
## Review — Kira

**Focus:** fencing correctness, race conditions, authorization

### Findings

**1. [CRITICAL] Generation-mismatch fencing is documented but not implemented**

The module doc (lines 13-14) states:

> Generation mismatch -> stop_volume (stale placement, fencing)

But `reconcile_once` never checks generations of already-running volumes. The code only checks membership in the desired set vs. running set — it never compares `placement_generation` of the desired volume against the generation stored in the running `VolumeProcess`.

This means: if a volume is live-migrated (new generation issued in Raft), the old hypervisor will keep serving the stale generation indefinitely as long as the volume remains in its desired set. This is a split-brain / data-corruption vector.

**Fix:** After the start loop, iterate running volumes that ARE in the desired map but whose `placement_generation` differs from the tracked generation in `VolumeMgr`. Stop and restart them. This requires `VolumeMgr` to expose the generation of running volumes (e.g., `pub fn running_generation(&self, volume_id: &str) -> Option<u64>`).

**2. [MEDIUM] `list_active` returns only IDs — no generation metadata**

`VolumeMgr::list_active()` returns `Vec<String>`. The reconciler has no way to detect generation drift even if it wanted to. The API needs to return `(volume_id, generation)` pairs for the reconciler to do generation-aware diffing.

**3. [LOW] `encryption_passphrase` stored as plaintext `String` in struct**

`StorageReconciler` holds `encryption_passphrase: String` for the entire process lifetime. Consider using `secrecy::SecretString` (or zeroize-on-drop) so the passphrase does not linger in memory or appear in debug output. Not blocking, but worth a follow-up.

**4. [LOW] `last_report` uses `std::sync::Mutex` in async context**

`self.last_report.lock().unwrap()` is called inside `reconcile_once` which is `async`. A `std::sync::Mutex` held across `.await` would block the executor, but here the lock is released before any `.await`, so this is safe in practice. However, if the code evolves, it could become a footgun. Consider `tokio::sync::Mutex` or keep the pattern as-is with a comment explaining the invariant.

**5. [INFO] Tests are mostly no-op due to private `VolumeMgr` internals**

`reconcile_stops_orphaned_volume` spawns a `sleep` process but cannot inject it into `VolumeMgr` (private `processes` field), so it asserts `stopped == 0` — not actually testing the stop path. `reconcile_reaps_dead_processes` similarly asserts 0 reaps. These tests document intent but do not validate behavior. Consider adding a `VolumeMgr::inject_test_process` (cfg(test)) or integration tests.

### Summary

The reconciler structure is solid — clean trait abstraction, proper shutdown handling, and the start/stop diff logic is correct for the membership check. However, the **generation fencing path is missing**, which is the security-critical piece for preventing split-brain on volume migration. The doc comment promises it but the code does not deliver.

**Verdict: REQUEST_CHANGES**

Must fix before merge:
- Implement generation-mismatch detection and restart (finding 1)
- Expose running generation from `VolumeMgr` (finding 2)

PR #1345, opened 2026-04-05
Key hypervisor store by ID with name-to-id index

### LINKED ISSUE #1316: ID migration: Hypervisor keyed by ID
Parent epic: #1312
Depends on: core ID types

## Task

Migrate Hypervisor from name-keyed to ID-keyed.

### Current state
- Hypervisor keyed by `name` in HypervisorStore
- All Raft commands use `name`: `RegisterHypervisor { name }`, `EnableHypervisor { name }`, `DrainHypervisor { name }`
- Scheduler references hypervisors by name
- Already has an `id` field (`hv-{ulid}`) but it's not used as the primary key

### Changes

1. **HypervisorStore**: Key by `HypervisorId`, name→id index
2. **Raft commands**: All use `HypervisorId` instead of name
3. **Scheduler**: `schedule()` returns `HypervisorId`
4. **Gossip**: Reports keyed by HypervisorId
5. **VmPlacement.hypervisor_id**: Already string, make typed
6. **Auto-register (daemon.rs)**: Generate ID at registration

## Acceptance criteria
- [ ] Hypervisor keyed by ID in Raft and redb
- [ ] Scheduler works with IDs
- [ ] `syfrah hypervisor list` shows both ID and name

sacha-ops:
## Summary
- **HypervisorStore** now uses `HypervisorId` as the primary key instead of `name`. A secondary `hypervisors_name_idx` table maps name to id for backward-compatible lookups.
- **HypervisorId::generate()** added to `types.rs` for unique ID creation (timestamp + process-id based).
- **RegisterHypervisor** Raft command generates proper unique IDs instead of `hv-{name}`.
- **HypervisorCandidate** in the scheduler gains an `id` field populated from the store or gossip report.
- **migrate_legacy_record()** method for transparent upgrade of old name-keyed entries.
- All existing tests pass, new tests for ID-based operations and legacy migration.

## Test plan
- [x] `cargo test -p syfrah-org --lib` passes (288 tests)
- [x] `cargo test -p syfrah-controlplane --lib` passes (165 tests)
- [ ] Verify gossip reports use the correct hypervisor ID in multi-node E2E

Closes #1316

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2983,6 +2983,7 @@ dependencies = [
  "console",
  "dirs",
  "ipnet",
+ "rand 0.8.5",
  "serde",
  "serde_json",
  "syfrah-api",
--- a/layers/compute/src/network_setup.rs
+++ b/layers/compute/src/network_setup.rs
@@ -696,7 +696,7 @@ mod tests {
     use std::sync::Arc;
 
     use syfrah_org::ipam::IpamStore;
-    use syfrah_org::types::{EnvironmentId, ProjectId, VpcOwner};
+    use syfrah_org::types::VpcOwner;
     use syfrah_org::PlacementStore;
     use syfrah_overlay::MockBackend;
     use tempfile::TempDir;
@@ -736,9 +736,11 @@ mod tests {
             org_store.create(ORG_NAME).unwrap();
             org_store.create_project(ORG_NAME, PROJECT_NAME).unwrap();
 
-            let project_id = ProjectId(format!("{ORG_NAME}/{PROJECT_NAME}"));
-            let env_id = EnvironmentId(format!("{ORG_NAME}/{PROJECT_NAME}/{ENV_NAME}"));
-            org_store
+            let project = org_store
+                .get_project(ORG_NAME, PROJECT_NAME)
+                .unwrap()
+                .unwrap();
+            let env = org_store
                 .create_env(
                     ORG_NAME,
                     PROJECT_NAME,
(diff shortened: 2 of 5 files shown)


sacha-ops:
## Review — Soren

**Verdict: APPROVE**

### Systems findings

**HypervisorStore migration:**
The migration logic in `migrate_legacy_record()` is clean and correct:
1. Idempotent — returns early if the name index already has an entry
2. Reads legacy name-keyed record from the primary table
3. Deletes old name-keyed entry, writes ID-keyed entry + name index
4. The `Hypervisor` struct already carries its `id` field (even legacy records had `hv-{name}` format), so no data is lost

The store API surface changes are backward-compatible: `get()` still takes a name, `get_by_id()` is additive, `delete()` still takes a name with proper index cleanup. The `update()` method correctly refreshes the name index entry in case of any inconsistency.

**Scheduler `HypervisorCandidate` ID field:**
The new `id` field is correctly populated in both construction paths:
- `from_hypervisor()`: uses `hv.id.0.clone()` — correct, reads from the store record
- `from_gossip_report()`: uses `report.hypervisor_id.clone()` — correct, reads from gossip

The test `hypervisor_candidate_utilization` is updated with a dummy ID. The scheduler itself doesn't need to key by ID yet (it still selects by scoring), but having the ID available on the candidate struct is the right prep for future placement tracking.

**State machine change:**
`state_machine.rs` switches from `HypervisorId(format!("hv-{name}"))` to `HypervisorId::generate()`. This is the critical fix — no more name-derived IDs that break when hypervisors are renamed or re-registered. Correct.

**Table names:**
`table_names()` now returns both `TABLE` and `NAME_INDEX`, ensuring snapshot export/import captures the index. Good.

Clean migration, no issues found.

PR #1348, opened 2026-04-05
Key VPC, Subnet, SG, NAT GW, Route Table by generated ID

### LINKED ISSUE #1315: ID migration: VPC, Subnet, Security Group keyed by ID
Parent epic: #1312
Depends on: Org ID migration

## Task

Migrate VPC, Subnet, Security Group, NAT Gateway, Route Table from name-keyed to ID-keyed.

### Current state
- VPC keyed by `name` globally
- Subnet keyed by `"vpc/subnet"` composite
- SG keyed by `"vpc_id/sg_name"` composite
- NAT GW, Route Table similar composites

### Changes

1. **VpcStore**: Key by `VpcId`, name→id index scoped to org
2. **SubnetStore**: Key by `SubnetId`, name→id index scoped to VPC
3. **SgStore**: Key by `SgId`, name→id index scoped to VPC
4. **NatGwStore, RouteTableStore**: Same pattern
5. **All Raft commands**: Switch from name to ID
6. **VPC Peering**: Key by PeeringId, reference VPCs by VpcId
7. **IPAM**: Subnet references use SubnetId

### Foreign key updates
- Subnet.vpc_id → VpcId (typed)
- SG.vpc_id → VpcId (typed)
- NIC.subnet_id → SubnetId (typed)
- Route targets → typed IDs

## Acceptance criteria
- [ ] All networking resources keyed by ID
- [ ] Cross-VPC references use VpcId
- [ ] CLI resolves names within scope (vpc name unique per org)

sacha-ops:
## Summary
- Migrate all network resources (VPC, Subnet, Security Group, NAT Gateway, Route Table) from name/composite-based to random-ID-based primary keys in redb
- Add name-to-ID index tables for backward-compatible CLI name lookups
- Update all typed foreign keys (NetworkInterface, VpcPeering, VpcAttachment, VmPlacement, RouteTarget) to use typed IDs instead of raw strings

## Details
Each resource now gets a random 12-hex-char ID at creation (e.g., `vpc-a1b2c3d4e5f6`). Name uniqueness is enforced via scoped index tables:
- `idx_vpc_name`: `{org_scope}/{vpc_name}` -> VPC ID
- `idx_subnet_name`: `{vpc_id}/{subnet_name}` -> Subnet ID
- `idx_sg_name`: `{vpc_id}/{sg_name}` -> SG ID
- `idx_nat_gw_name`: `{vpc_id}/{nat_name}` -> NatGw ID
- `idx_route_table_name`: `{vpc_id}/{table_name}` -> RouteTable ID

All downstream layers (controlplane, compute, fabric, overlay) updated to work with the new typed IDs.

## Test plan
- [x] `cargo test --workspace` passes (544/545, 1 pre-existing failure in binary.rs)
- [ ] E2E: Create VPC, Subnet, SG, NAT GW, Route Table and verify IDs are generated
- [ ] E2E: Verify name-based lookups still work after migration

Closes #1315

### DIFF
--- a/layers/compute/src/cli/vm.rs
+++ b/layers/compute/src/cli/vm.rs
@@ -75,6 +75,11 @@ pub enum VmCommand {
         /// Single-node: warning emitted.
         #[arg(long)]
         spread_topology: Option<String>,
+        /// Wait for the VM to reach Running or Failed phase before returning.
+        /// Without this flag, the CLI returns immediately after the intent is
+        /// accepted (phase=Pending).
+        #[arg(long)]
+        wait: bool,
     },
     /// List all virtual machines
     #[command(after_help = "Examples:\n  syfrah compute vm list\n  syfrah compute vm list --json")]
@@ -157,6 +162,7 @@ pub async fn run(cmd: VmCommand) -> anyhow::Result<()> {
             node_selector,
             anti_affinity,
             spread_topology,
+            wait,
         } => {
             run_create(
                 name,
@@ -176,6 +182,7 @@ pub async fn run(cmd: VmCommand) -> anyhow::Result<()> {
                 node_selector,
                 anti_affinity,
                 spread_topology,
+                wait,
             )
             .await
         }
@@ -294,6 +301,7 @@ async fn run_create(
     node_selector: Vec<String>,
     anti_affinity: Option<String>,
     spread_topology: Option<String>,
+    wait: bool,
 ) -> anyhow::Result<()> {
     let ssh_key = match ssh_key_path {
         Some(ref path) => Some(read_ssh_key(path)?),
@@ -313,7 +321,7 @@ async fn run_create(
     };
 
     let req = ComputeRequest::CreateVm {
-        name,
+        name: name.clone(),
         vcpus,
         memory_mb: memory,
         image,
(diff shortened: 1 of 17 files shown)


sacha-ops:
## Review — Nova

**Verdict: APPROVE**

Reviewed the typed ID conversions for unnecessary allocations. Findings:

**`.0` field access is zero-cost.** The newtype pattern (`VpcId(String)`, `SubnetId(String)`, etc.) means `.0` is a direct field access — no allocation, no copy. The ~30+ sites in `daemon.rs` and `state_machine.rs` doing `p.vpc_id.0.clone()` are allocating, but they were already allocating with the old `p.vpc_id.clone()` when `vpc_id` was a bare `String`. Net-zero allocation change.

**`generate_id` uses `rand::thread_rng` correctly.** The per-character `rng.gen::<u8>() % 16` approach generates one random byte to extract one hex digit, which wastes 4 bits per char. A more efficient approach would be to generate 6 random bytes and hex-encode them directly. However, ID generation happens once per resource creation — this is not a hot path. The overhead is ~12 random bytes instead of 6 per call. Negligible.

**List-then-filter pattern for VPC-scoped queries.** Methods like `list_subnets_by_vpc_id`, `list_sgs_by_vpc`, `list_route_tables_by_vpc`, and `list_nat_gws_by_vpc` now load all entries and filter by `vpc_id` field match, replacing the old prefix-scan approach. Both are O(n) full table scans via `db.list()`, so the performance characteristic is unchanged. The filter predicate now compares `VpcId` structs (derived `PartialEq` on String) instead of string prefix matching — same cost.

**Peering operations do redundant full scans.** `create_peering` calls `list_peerings()` for duplicate detection, `delete_peering` calls it again for finding the target, and `get_peering` does the same. Each call deserializes every peering record. For the typical peering count (< 50), this is sub-millisecond. If peering counts grew to thousands, a secondary index (like the name indexes for other resources) would help. Not worth optimizing now.

**`get_vpc()` suffix scan allocates the full index.** Loading all `VPC_NAME_INDEX` entries into a `Vec<(String, String)>` to find one by suffix is allocation-heavy compared to a direct key lookup. But this only happens for name-based VPC lookups (CLI convenience paths), not ID-based lookups which are the primary internal path. The `get_vpc_by_id()` method is O(1) and allocation-minimal.

**No hot-path regressions.** The overlay/placement code in `daemon.rs` converts `VpcId` → `String` via `.0.clone()` which is exactly the same allocation as before. The data plane (FDB updates, VXLAN forwarding) does not touch these ID types at all — it operates on the already-extracted strings in `syfrah_overlay::VmPlacement`.

No performance concerns. Clean migration with no allocation regressions.

PR #1350, opened 2026-04-05
Add CLI name-to-ID resolution layer

### LINKED ISSUE #1318: ID migration: CLI name-to-ID resolution layer
Parent epic: #1312
Depends on: all resource ID migrations

## Task

Add a name resolution layer in the CLI so operators use names but the system works with IDs.

### Design

```rust
// Before sending any command, resolve names to IDs
let org_id = resolve_org(name)?;           // "acme" → org-a1b2c3d4
let project_id = resolve_project(org_id, name)?;  // "backend" → proj-e5f6g7h8
let vpc_id = resolve_vpc(org_id, name)?;   // "my-vpc" → vpc-m3n4o5p6
```

### Changes

1. **`layers/*/src/cli/`**: Every CLI command resolves names before sending
2. **Control socket protocol**: Commands accept IDs (not names)
3. **Ambiguity handling**: If name matches multiple resources, error with list
4. **ID pass-through**: If user passes an ID directly (`vm-a1b2c3d4`), skip resolution
5. **Tab completion**: Complete from names, send IDs

### UX

```bash
# Both work:
syfrah compute vm get web-1           # resolves name → ID
syfrah compute vm get vm-a1b2c3d4     # uses ID directly

# Ambiguity:
syfrah vpc delete my-vpc
Error: multiple VPCs named 'my-vpc':
  vpc-m3n4o5p6 (org: acme)
  vpc-q7r8s9t0 (org: other)
Use --org or pass the VPC ID directly.
```

## Acceptance criteria
- [ ] All CLI commands resolve names → IDs
- [ ] IDs accepted directly (skip resolution)
- [ ] Clear error on ambiguous names
- [ ] No change to user-facing UX (names still work)

sacha-ops:
## Summary

- Adds a `resolve` module to `syfrah-core` with `ResourceType` enum, `looks_like_id()` prefix detection, and `ResolveResult`/`ResolveMatch` types for structured resolution outcomes
- Adds `OrgRequest::Resolve` / `OrgResponse::Resolved` protocol variants for generic name-to-ID resolution over the daemon control socket
- Implements daemon-side resolution handler covering: org, project, env, vpc, subnet, sg, hypervisor, nat-gw, and route-table -- with ambiguity detection that returns all matches with disambiguation context
- Adds `resolve_name()` async client-side helper that CLI commands can call -- handles ID pass-through, single-match resolution, and formatted ambiguity errors
- Registers `Resolve` as a read-only request in the Raft handler so it can be served by any node

### ID pass-through

If input starts with the resource prefix (e.g., `vpc-`), it is returned as-is without contacting the daemon.

### Ambiguity handling

When a name matches multiple resources, the error includes all matches with context:
```
multiple VPCs named 'web':
  vpc-a1b2c3d4 (org: acme)
  vpc-e5f6g7h8 (project: other/backend)
Use the appropriate scope flag to disambiguate, or pass the VPC ID directly.
```

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo test -p syfrah-core -- resolve` -- 7 tests pass (unit + doctest)
- [x] `cargo test -p syfrah-org` -- 289 tests pass including doctest
- [ ] Integration: verify `syfrah vpc delete my-vpc` resolves name to ID before sending to daemon

Closes #1318

### DIFF
--- a/layers/compute/src/manager.rs
+++ b/layers/compute/src/manager.rs
@@ -879,8 +879,8 @@ impl VmManager {
                         if let Err(te) = ns
                             .teardown(
                                 &vm_id_str,
-                                &nr.placement.vpc_id,
-                                &nr.placement.subnet_id,
+                                &nr.placement.vpc_id.0,
+                                &nr.placement.subnet_id.0,
                                 &nr.subnet_cidr,
                                 &nr.ip,
                                 &nr.tap_name,
@@ -949,8 +949,8 @@ impl VmManager {
         // Extract network metadata from setup result for runtime state.
         let (vm_ip, vm_subnet, vm_vpc, net_info) = if let Some(ref nr) = network_result {
             let info = NetworkInfo {
-                vpc_id: nr.placement.vpc_id.clone(),
-                subnet_id: nr.placement.subnet_id.clone(),
+                vpc_id: nr.placement.vpc_id.0.clone(),
+                subnet_id: nr.placement.subnet_id.0.clone(),
                 subnet_cidr: nr.subnet_cidr.clone(),
                 ip: nr.ip.clone(),
                 mac: nr.mac.clone(),
@@ -965,7 +965,7 @@ impl VmManager {
                         .map(|s| s.name.clone())
                         .unwrap_or_default(),
                 ),
-                Some(nr.placement.vpc_id.clone()),
+                Some(nr.placement.vpc_id.0.clone()),
                 Some(info),
             )
         } else {
@@ -1049,7 +1049,7 @@ impl VmManager {
                     Arc::clone(backend),
                     local_node.clone(),
                 );
-                if let Err(e) = ns.mark_assigned(&nr.placement.subnet_id, &nr.ip, &vm_id_str) {
+                if let Err(e) = ns.mark_assigned(&nr.placement.subnet_id.0, &nr.ip, &vm_id_str) {
                     warn!(
                         vm_id = %vm_id_str,
                         error = %e,
--- a/layers/compute/src/network_setup.rs
+++ b/layers/compute/src/network_setup.rs
@@ -12,8 +12,8 @@ use tracing::{info, warn};
 use syfrah_org::ipam::IpamStore;
 use syfrah_org::store::OrgStore;
 use syfrah_org::types::{
-    NetworkInterface, NicId, PlacementAction, ResourceState, SecurityGroupId, Subnet, VmPlacement,
-    Vpc,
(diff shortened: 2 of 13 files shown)


sacha-ops:
## Review — Milo

**Verdict: APPROVE**

Reviewing resolution UX, ambiguity handling, and ID pass-through.

### ID pass-through

The `looks_like_id` function (resolve.rs:123) checks if the input starts with `{prefix}-`. If so, the CLI and the daemon both short-circuit and return the input as-is without a socket call. This is clean — users who already have an ID never hit the daemon. The client-side `resolve_name` (api.rs:725) duplicates this check before the socket call, which is a good optimization.

### Ambiguity errors

When multiple resources match a name, the `Ambiguous` variant provides a list of `ResolveMatch` structs with `id` and `context` (e.g., "org: acme", "vpc: production"). The client-side error message (api.rs:743-751) formats this clearly:

```
multiple vpcs named 'web':
  vpc-aaa (org: acme)
  vpc-bbb (org: other)
Use the appropriate scope flag to disambiguate, or pass the vpc ID directly.
```

This is helpful — it tells the user exactly what matched, where, and what to do next. The `context` field correctly varies by resource type (org for VPCs, VPC for subnets/SGs).

### Resolution UX by resource type

- **Org**: Direct name lookup, no scope needed. Simple and expected.
- **Project**: Requires `--org` scope. Error message is clear: "project resolution requires --org scope".
- **Env**: Requires scope in "org/project" format. Error messages for missing/malformed scope are specific.
- **VPC**: Tries direct name lookup first, falls back to org-scoped scan, then global scan. Progressive resolution — nice UX.
- **Subnet**: Scoped by VPC name. Without scope, scans all VPCs and reports ambiguity. Correct.
- **SG**: Same pattern as subnet — VPC-scoped with global fallback.
- **Hypervisor**: Direct name lookup via hv_store.
- **NatGw, RouteTable**: Global scan with ambiguity detection.
- **Vm, Volume, Snapshot**: Correctly deferred to their own layers with a clear error message.

### Minor observation

The `looks_like_id` check only validates the prefix pattern (`{prefix}-`), not the hex suffix. This means `vpc-my-name` would pass through as an ID. This is intentional and documented in the function — it avoids false negatives where a valid ID might have an unexpected format. If the ID turns out to be invalid, the downstream lookup will return NotFound. Acceptable trade-off.

### Test coverage

The resolve module has good unit tests: prefix detection, label roundtrip, Display formatting, serde roundtrip. Integration with the daemon handler is tested implicitly through the OrgRequest::Resolve match arm.

Clean, user-friendly resolution layer. No issues.

